PR #53, opened 2015-05-20
Bar chart improvements

lomereiter:
General:
* MANY bug fixes
* Bar chart is automatically updated (with a nice d3.js transition) when
    * checkboxes are checked/unchecked
    * values of the trait are edited
    * buttons from the 'Block samples' panel are clicked

For samples with case attributes:
* Legend is now positioned at the top left above the plot
* If there are < 10 distinct values of an attribute, it looks like this:
legend1
* Otherwise the attribute is assumed to be continuous, and the old legend is shown:
legend2


zsloan:
Awesome, thanks. This is much better.

PR #76, opened 2015-06-20
fix mapping results loading

### LINKED ISSUE #71: Can't do mapping more than once without refreshing the trait page
This is a really embarassing issue which confuses the user, so I'm aiming to fix this today. If there isn't an easy/good way to fix this while still using modals, I'm just going to switch to opening a new window instead.

### LINKED ISSUE #74: Need to redo submission of mapping methods to not use ajax
I think we decided to use ajax so we could have a working progress bar on the trait page, but it seems that using ajax like this completely disables all javascript imports on the page afterwards. This is why it wouldn't map more than once without refreshing (specifically, it would throw an error when it attempted to create the loading bar modal because it no longer had the necessary bootstrap import). So not only could you not map again, but all other javascript on the page that uses jquery (or any other import, like bootstrap) would stop working (for example, the "block samples" option).

From the research I've done on this, a workable solution doesn't seem to exist. If I'm wrong about this by all means let me know, but I've come across a number of online posts where no one is able to give a solution. I think that I'm going to just have to submit the form normally and have the loading bar be a part of the destination (mapping results) page.

lomereiter:
This fixes #71 & #74.

Now the html loaded via ajax doesn't contain links to external CSS/JS files (they are all referenced from the `show_trait.html` template instead), because that caused the JS imports breakage. (I still don't understand what happens under the hood when an ajax-loaded page references JS files, but apparently it's a big no-no.)


zsloan:
Thanks a lot for this fix; I thought that the problem was just inherently caused by loading in html via ajax, not just the html being loaded linking to external css/js files. This definitely saved a lot of time that might have been spent implementing a worse fix.

Issue #44, opened 2015-05-14
Look for unsafe SQL queries


pjotrp:
Before going live we should go through SQL queries that get modified from forms - they may appear as security breaches.

pjotrp:
Referring to this for Genenetwork1 https://github.com/genenetwork/genenetwork/issues/2. Marjan, can you have a look?

pjotrp:
It can be that sqlalchemy takes care of some it.

http://stackoverflow.com/questions/6713715/escaping-special-characters-using-sqlalchemy

PR #77, opened 2015-06-21
SQL security fixes

lomereiter:
I tried to not miss anything important, but still consider it a good idea for someone else to go through the results of `grep execute` before closing https://github.com/genenetwork/genenetwork2/issues/44
Notice that I left many non-parameterized queries as is, as they seem harmless (arguments for string formatting come from other queries and not from forms).


mamarjan:
Is there any reason why it's not mandatory on this project to use the explicit escaping mechanism provided by sqlalchemy for all queries? Shouldn't be so hard. And I wouldn't expect performance issues, or any other issues with such a rule.

lomereiter:
Marjan,
That would be great, but I think it'd be even better to use high-level SQLAlchemy features (such as [reflection](http://docs.sqlalchemy.org/en/latest/core/reflection.html)), that would improve readability as well.

PR #81, opened 2015-06-23
Fix SVG/PDF exporting

### LINKED ISSUE #80: Bug: Select download PDF breaks for single chromosome
After interval mapping, when I select one chromosome view of the QTL map the PDF download fails with an Internal Server Error. PDF does work for the full map so it is probably something minor (SVG may be working).

lomereiter:
* Fixed #80 (handlers were attached to buttons before they are loaded via ajax)
* Added CairoSVG module to `misc/requirements.txt`
* Fixed a couple of typos (attchment → attachment)
* Instead of drawing a text link on SVG for returning to full view, introduced a real button — mainly because otherwise this text is exported too, which is undesirable


lomereiter:
Hi, I've updated the PR. Please merge it while it can be done automatically ;)

Issue #119, opened 2015-09-19
WGCNA integration


DannyArends:
Creating an issue so we can put some idea's an discussion on Github 
My docker image is up2date (we needed an additional python package), New install script for R/qtl and WGCNA are found at: https://github.com/DannyArends/gn2-docker/blob/master/Renv_setup.R

I made an initial prototype for WGCNA in genenetwork2 (https://github.com/DannyArends/genenetwork2/tree/WGCNAintegration)
The current addition to genenetwork2:
 - Adds a button to the traits collection page
 - On press it loads WGCNA in the R environment
 - Displays an output message which traits were selected

Oww and please assign this issue to me..

robwwilliams:
Danny: Great progress. Love to get your ideas about how to develop the
options and interfaces linking GN and WGCNA. Steve or Peter may be able to
help if we get snagged on any problems or if we just want advice on good
default parameters.

On Sat, Sep 19, 2015 at 12:40 PM, Danny Arends <notifications@github.com>
wrote:

> Creating an issue so we can put some idea's an discussion on Github
> My docker image is up2date (we needed an additional python package), New
> install script for R/qtl and WGCNA are found at:
> https://github.com/DannyArends/gn2-docker/blob/master/Renv_setup.R
>
> I made an initial prototype for WGCNA in genenetwork2 (
> https://github.com/DannyArends/genenetwork2/tree/WGCNAintegration)
> The current addition to genenetwork2:
>
>    - Adds a button to the traits collection page
>    - On press it loads WGCNA in the R environment
>    - Displays an output message which traits were selected
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/genenetwork/genenetwork2/issues/119>.
>



-- 
Rob

Robert W. Williams, Ph.D.
UT-ORNL Governor's Chair in Computational Genomics
Chair, Department of Genetics, Genomics and Informatics
University of Tennessee Health Science Center
Room 501
855 Monroe Avenue, Memphis TN 38163 USA

Office 901 448-7018  CELL 901 604 4752
Office: 501 Wittenborg Building
Department of Genetics: 71 Manassas St, Memphis TN 38163
EMAIL: rwilliams@uthsc.edu
Alternative email: labwilliams@gmail.com
SKYPE: robwwilliams

DannyArends:
WGCNA update:

We can run the WGCNA analysis on a selected collection.

TODO'S:
[...]

PR #120, opened 2015-10-08
WGCNA integration

DannyArends:
Hey all,

The inital WGCNA code has been pretty much finished, and I think we should start testing it on a real GN2 server using some hand picked testers / users.

Todo:
- Add more input fields for parameters (such as: dendroLabels, hang, etc) and pass them to WGCNA
- Buttons on the result-page should work (!!) Zach can you look into this ? The buttons redo with power on the results page should submit back to wgcna_results.html, the eigen gene buttons should add the data to the collection (and return to that screen)
- Need to add additional steps for follow up analysis per Module (GO, whatever)
- Tables should be downloadable in CSV

Some known issues: 
- Buttons on the results page do not work yet, so at the moment ONLY softThresholdSelection is possible (you cannot force it to use a specific power).
- I have no idea how to add the eigen genes to the current collection ( I just print them to screen )
- If you select 10k + genes :) it will be sllloooowwww


DannyArends:
Related to issue #119

PR #189, opened 2016-09-26
Changed the URL parameter from an int to a string

lyan6:


pjotrp:
Your branch contains conflicts with the master branch. Please resolve these first. Also roll back your master branch by cloning from the main tree again (see my E-mail). We'll rebase next.

Issue #191, opened 2016-09-28
Global search "il6" for "Internal Server Error"


lyan6:
When searching "il6" with global search, got an error.

On browser:
Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.

On server shell:
ProgrammingError: (ProgrammingError) (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'H3|2 87.58 cM  AND\n                                    InbredSet.Name = 'BXD' AN' at line 6") "\n                            SELECT\n                                    HomologeneId\n                            FROM\n                                    Homologene, Species, InbredSet\n                            WHERE\n                                    Homologene.GeneId = 2 H3|2 87.58 cM  AND\n                                    InbredSet.Name = 'BXD' AND\n                                    InbredSet.SpeciesId = Species.Id AND\n                                    Species.TaxonomyId = Homologene.TaxonomyId\n                            " ()
ERROR:werkzeug:Error on request

lyan6:
I fixed this issue in my branch. But after I did these:
cd gn2
git checkout master
git checkout -b 5ad7e4634f2e8b5cc5b10442da2dd80385e9aee6 5ad7e4634f2e8b5cc5b10442da2dd80385e9aee6
git branch -D master
git checkout -b master
git pull  git@github.com:genenetwork/genenetwork2.git 

I got some Unmerged files:
# Unmerged paths:
#   (use "git add/rm <file>..." as appropriate to mark resolution)
#
#       deleted by us:      genotype_files/genotype/BXD.geno
#       deleted by us:      genotype_files/genotype/MDP.geno
#       deleted by us:      genotype_files/genotype/json/AXBXA.json
#       deleted by us:      genotype_files/genotype/json/B6D2RI.json
#       deleted by us:      genotype_files/genotype/json/BXD.json
#       deleted by us:      genotype_files/genotype/json/HSNIH.json
#       both modified:      wqflask/wqflask/show_trait/show_trait.py

So I can resolve these things, and commit/push anyway?

@pjotrp @zsloan

pjotrp:
Yes, please push to your own repo. 

Also use GENENEWORK_FILES to point to the genotype dir from now on: https://github.com/genenetwork/genenetwork2/blob/master/etc/default_settings.py#L56. The genotype files should no longer be in the source tree.

lyan6:
[...]

PR #192, opened 2016-09-29
Fix global search "il6" for "Internal Server Error". Add single quote…

lyan6:
…s into MySQL query.

https://github.com/genenetwork/genenetwork2/issues/191


pjotrp:
Thanks!

PR #201, opened 2016-10-04
Changes separating R/qtl mapping code from marker_regression.py

zsloan:
This commit also contains a minor improvement to the appearance of the Mapping Tools form.


pjotrp:
Thanks! I'll merge it after removing the genotype selector and running regression tests.

PR #203, opened 2016-10-05
CTL now displays proper error if fewer than 4 traits are selected. 

zsloan:
Also, WGCNA previously incorrectly gave an error when fewer than 5 traits were selected instead of fewer than 4 (because it said '<= 4').

I also improved the formatting for the html for both CTL and WGCNA Analysis.


pjotrp:
Thanks for splitting the files up! With CTL the minimum is 2 traits - I'll change that.

pjotrp:
I also changed the word phenotypes into traits.

PR #208, opened 2016-10-07
Removed qtlreaper code from marker_regression and moved it to a separ…

zsloan:
…ate file.

Fixed CTL template to display correct number of traits (previously it would display 1 fewer than you selected).


pjotrp:
Looking good. I'll merge and test tomorrow.

PR #210, opened 2016-10-12
Removed more code from marker_regression

zsloan:


wqflask/wqflask/collect.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, collection_name):
         self.id = None
         self.created_timestamp = datetime.datetime.utcnow().strftime('%b %d %Y %I:%M%p')
         self.changed_timestamp = self.created_timestamp #ZS: will be updated when changes are made
-
+
```
pjotrp: Whitespace inserted

wqflask/wqflask/marker_regression/marker_regression.py
```diff
@@ -212,7 +212,8 @@ def __init__(self, start_vars, temp_uuid):
                                                                                                                                                         self.control_marker,
```
pjotrp: Why these empty lines?

wqflask/wqflask/marker_regression/plink_mapping.py
```diff
@@ -0,0 +1,168 @@
...
+                                                                                                                                                                                                                    PLINK_PATH, 
+                                                                                                                                                                                                                    dataset.group.name, 
+                                                                                                                                                                                                                    TMPDIR, 
+                                                                                                                                                                                                                    plink_output_filename, 
+                                                                                                                                                                                                                    this_trait.name,
```
pjotrp: lines

wqflask/wqflask/marker_regression/plink_mapping.py
```diff
@@ -0,0 +1,168 @@
...
+    count, p_values = parse_plink_output(plink_output_filename, species)
+
+    #for marker in self.dataset.group.markers.markers:
+    #    if marker['name'] not in included_markers:
+    #        logger.debug("marker:", marker)
+    #        self.dataset.group.markers.markers.remove(marker)
+    #        #del self.dataset.group.markers.markers[marker]
+
+    logger.debug("p_values:", pf(p_values))
+
```
pjotrp: Extraneous ^M injected in new file (Windows-style)

zsloan:
Moving the parameters to new lines was to avoid it making the line way too
long and running off the edge of the screen (in the case of plink_command
it's already really long even before doing that). How else should I deal
with that?

Also, where is the ^M in plink_mapping? My editor (notepad++) says it
should be using Unix End of Line conversion. I also told my editor to show
me the end of line characters and couldn't find any ^M's.

leiyan:
Zach, maybe you can install the “Hex Editor Plugin” into Notepad++, then you can know ASCII codes.

zsloan:
I installed the plugin Lei mentioned, but wasn't able to find the hex for
^M (0d) anywhere in the file.

pjotrp:
Hi Zach, the problem is that it is a Windows file. There is some hidden magic. Last time I converted it to unix with vim for you using :set ff=unix. See http://stackoverflow.com/questions/82726/convert-dos-line-endings-to-linux-line-endings-in-vim. 

As we deploy on Linux we have to use Unix-style line endings. Just check the PR and see if it looks OK to you on github. If it does I'll accept it. Github is Unix-style. There is also a git option (I remember now) which may help http://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows

zsloan:
Very weird, because the file was set to Unix when I checked in the editor.
Maybe for some reason it didn't affect the file retroactively.

I'm still a little confused about where the ^M is; are you seeing it in the
code snippet that you linked from the pull request (and if so, where
specifically)? The hex for the ^M line break doesn't show up anywhere in
the file and I don't see it in the (previous) pull request.

Also, could you explain what I should do regarding the function parameters?
Is it possible that you didn't see them to the right (you have to scroll I
think) and assumed they were just random empty lines? This applies to both
the second and third comments you made. I don't mind changing that back to
how it was, but I thought it was a problem to have those two lines run
really far off the right edge of the page.

wqflask/wqflask/marker_regression/plink_mapping.py
pjotrp: yes, this is inlined and the github viewer won't show it. Better format it in a way that you don't exceed 80 cols - or at least have the next line start before col 80.

pjotrp:
I'll merge this PR and make the changes for you. Next time use the github PR files viewer to see how it pans out.

pjotrp:
Merged. This was what I needed to do to correct this PR: https://github.com/genenetwork/genenetwork2/commit/a734338acd10c015e35da9574c8ef0b9dee68704

zsloan:
Oh, is it possible that the thing you thought was inserted ^M's was
actually just me spacing poorly? Because I think I might have added those
spaces on purposes for some dumb reason.

PR #216, opened 2016-10-25
Replaced Download Table in view_collection page with python version

zsloan:
Also added more trait parameters to the jsonable function in GeneralTrait in anticipation of implementing the DataTables Scroller feature using client-side processing (of a json version of the table's contents)


pjotrp:
Thanks!

PR #219, opened 2016-10-29
For those case that we have multiple genotypes we would like to add a…

lyan6:
… selector.


pjotrp:
I merged in testing. Please do a PR request against testing next time.

PR #234, opened 2016-11-15
Various changes

zsloan:
If you want to hold off on merging the Scroller stuff, then just merge every commit except these two:
717677b0c09f6ba08268db12d4889503cc2606d9
7b83c8c46977942acb271e7f1cf187d00bc7ff5a


pjotrp:
@zsloan is there a reason not to merge the scroller stuff?

zsloan:
I just wasn't sure if I should, given how it's kind of kludgy (and
currently only implemented for the regular search result page as a result;
I can add it to other tables but it will require duplicating some code in
each of them). I think it should work okay, though.

pjotrp:
It is kinda hard to asses what to do ;). Next time, if you put in a PR, it should be safe for me to do so. That is the main assumption - I can only read the code and do some basic tests. I'll merge this tomorrow.

wqflask/base/webqtlCaseData.py
```diff
@@ -66,7 +66,7 @@ def class_outlier(self):
 
     @property
     def display_value(self):
-        if self.value:
+        if self.value != None:
```
pjotrp: Should we not use self.value is not None: Comparisons to singletons like None should always be done with
'is' or 'is not', never the equality operators.

wqflask/base/trait.py
```diff
@@ -438,6 +507,18 @@ def retrieve_info(self, get_qtl_info=False):
                     self.location_repr = 'Chr%s: %.6f' % (self.chr, float(self.mb))
                     self.location_value = trait_location_value
 
+            elif self.dataset.type == "Geno":
+                self.location_repr = 'N/A'
+                self.location_value = 1000000
```
pjotrp: Can we move all state variables into a separate object? The use of self is kinda dangerous and a code smell in my opinion.

wqflask/base/trait.py
```diff
@@ -109,6 +122,62 @@ def jsonable(self):
         else:
             return dict()
 
+    def jsonable_table_row(self, index):
```
pjotrp: Can the method have a better name? If it does not return json...

wqflask/base/trait.py
```diff
@@ -109,6 +122,62 @@ def jsonable(self):
...
+                return ['<input type="checkbox" name="searchResult" class="checkbox trait_checkbox" style="transform: scale(1.5);" value="' + user_manager.data_hmac('{}:{}'.format(str(self.name), self.dataset.name)) + '">',
+                        index,
+                        '<a href="/show_trait?trait_id='+str(self.name)+'&dataset='+self.dataset.name+'">'+str(self.name)+'</a>',
+                        self.description_display,
+                        self.authors,
+                        self.pubmed_text,
+                        self.LRS_score_repr,
+                        self.LRS_location_repr,
+                        additive]
+        elif self.dataset.type == "Geno":
```
pjotrp: The use of direct HTML is a code smell. How can we avoind that?

wqflask/base/trait.py
```diff
@@ -459,7 +540,7 @@ def retrieve_info(self, get_qtl_info=False):
                     logger.sql(query)
                     trait_qtl = g.db.execute(query).fetchone()
                     if trait_qtl:
-                        self.locus, self.lrs, self.pvalue, self.mean, self.additive= trait_qtl
+                        self.locus, self.lrs, self.pvalue, self.mean, self.additive = trait_qtl
```
pjotrp: Initialization of local state (self.) here is pretty scary! Do we really need to maintain state in this object?

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -216,7 +216,7 @@ def __init__(self, start_vars):
             for _trait_counter, trait in enumerate(self.correlation_data.keys()[:self.return_number]):
                 trait_object = GeneralTrait(dataset=self.target_dataset, name=trait, get_qtl_info=True, get_sample_info=False)
 
-                if self.dataset.type == "ProbeSet" or self.dataset.type == "Geno":
+                if self.target_dataset.type == "ProbeSet" or self.target_dataset.type == "Geno":
```
pjotrp: target_dataset is a new variable. Do we really need that? It is getting confusing...

wqflask/wqflask/templates/correlation_matrix.html
```diff
@@ -26,35 +27,53 @@
       <td align="center" rowspan="{{traits|length + 1 }}" width="10" style="font-weight: Bold; border: 1px solid #999999; padding: 3px; color: #fff; background-color: royalblue;">P e a r s o n &nbsp;&nbsp;&nbsp; r</td>
```
pjotrp: Can we start using CSS for these HTML attributes? This should really be a css class!

PR #244, opened 2017-02-08
Merge Biodalliance work & coffeescript removal

chfi:
Contains the Biodalliance stuff and the removal of the coffeescript files. Did a final compilation on the coffeescript files that were consistent with the JS because a newer version of the CS compiler has prettier output...
Dropping both into the same PR because I'm lazy.


pjotrp:
Thanks Chris. Note I can not merge right now because there is a mixture with one of my branches and my commits. When I get round to merging I'll include this.

pjotrp:
@chfi we'll have to redo this work for the new version. Better not merge, but keep a copy around.

PR #255, opened 2017-06-30
Correlation Scatterplot update

lyan6:


pjotrp:
This pull request contains commits from other trees than master. I am not sure this is intentional, though it looks harmless enough. @zloan please check this PR separately in a new branch and see if it works for you. If not, @lyan6 should provide a clean PR against our development branch - that is the one we are working on. The master branch reflects what is running on 'production'.

zsloan:
Hm, so should this pull request maybe be made to the development or testing branches instead (I think development is the right one)? I'm guessing Lei would have to redo the pull request and there's no way for me to "redirect" this to a different branch?

PR #270, opened 2017-12-20
Merge some older commits

pjotrp:
Nothing too serious. But it may override some environment settings. So look at these if you face problems after the merge.


pjotrp:
@zsloan can you please merge.

PR #275, opened 2018-01-31
Added cli swith and set timeout to 3 minutes for gunicorn-dev mode

pjotrp:


zsloan:
Should I try pulling and running this on my branch?

Issue #135, opened 2016-01-21
Removing source packages from GN


pjotrp:
@zsloan Some large packages here in the source tree. Which ones are actually in use?

1.1M    bootstrap
404K    bootstrap-old
4.7M    ckeditor
1.4M    colorbox
16K     DT_bootstrap
7.6M    jqplot
104K    node
4.2M    novus-nvd3-d51729c
492K    pragmaticly-smart-time-ago-40fa376
0       smart-time-ago
2.0M    TableTools
20K     underscore

This relates to https://github.com/genenetwork/genenetwork2/issues/134

pjotrp:
We have a route to do this now - the way we are deploying biodalliance.

pjotrp:
- [ ] bootstrap
- [ ] ckeditor
- [ ] colorbox
- [ ] cytoscape
- [ ] DT_bootstrap
- [ ] jqplot
- [ ] pragmaticly-smart-time-ago-40fa376
- [ ] smart-time-ago
- [ ] TableTools
- [ ] underscore
- [ ] DataTables
- [ ] noUiSlider
- [ ] nvd3
- [ ] ValidationPlugin
- [ ] chroma.js
- [ ] d3-tip.min.js
- [ ] jquery.cookie.js
- [ ] json2.js
- [ ] jstat.min.js
- [ ] jszip.min.js
- [ ] parsley.min.js
- [ ] plotly-latest.min.js
- [ ] shapiro-wilk.js
- [ ] underscore-min.js
- [ ] underscore.string.min.js
- [ ] zxcvbn

PR #301, opened 2018-04-02
Testing rm bloat (jqplot and ckeditor)

pjotrp:
Hi Zach, jqplot and ckeditor are not used, so we should remove them. ckeditor pretends to edit help pages etc., but it is not functional. 

Javascript packages should come in from Guix (like we do with the twitter feed). I am going to move the ones we use today across. Trying to move all non-source-code out of our tree.


pjotrp:
We'll go one at a time with some time for testing. Not to confuse things too much. If anyone shouts I will re-add a package through Guix. Guix makes it much easier to deal with versioning. Besides it is more secure (an immutable resource).

pjotrp:
See also #135

zsloan:
I think this is okay; I also can't find anywhere that should need jqplot or ckeditor (I think the former was for some plot that may have been converted to plotly or the library Lei is using for the correlation scatterplot, and the latter may have been for editing html pages or something but doesn't seem to be in use now). Should I merge to testing and try it on my branch (and then staging if it seems to not be causing a problem there)?

pjotrp:
Yup.

PR #302, opened 2018-04-03
@acenteno added data upload scripts into main repo

pjotrp:


zsloan:
We already have a "gene/wqflask/maintenance" directory with some python scripts (like the one that generates BIMBAM from .geno files or the one that generates kinship matrices from BIMBAM files) so I'm wondering if either those files should go in this new directory or the other way around.

pjotrp:
Hi Zach, scripts not part of the webserver should be in the new directory. On difference between the maintenance scripts and the upload scripts is that the first needs the GN2_PROFILE and startup settings. But I expect the uploads scripts also will start using that in the future - simply because it is controlled dependencies. So, yes you can move the maintenance scripts into ./scripts. I will do it when I get there. I am cleaning up the webserver so it gets much easier to grep things and simplify.

zsloan:
Alright, I'll go ahead and merge the changes then.

PR #309, opened 2018-04-06
Testing rm bloat

### LINKED ISSUE #308: Move cytoscape.js out of tree

pjotrp:
Cytoscape and related removed and rejigged. Please test with GN2_PROFILE=/usr/local/guix-profiles/gn2-latest  and close #308 when it is OK.


pjotrp:
@zsloan make sure the network graph still works as usual. I tested it, so it should. Also the JS log did not complain. I mean there are many errors, but that is because jquery is out of date. I'll fix that soon.

PR #388, opened 2020-05-05
Add missing columns to db downloaded from AWS

BonfaceKilz:
#### Description

The zip file from the AWS S3 bucket has some missing columns, and this causes GN2 to throw assert errors because of missing columns. This PR adds the missing columns

  *Scenario: A developer sets up sql using the sql zip file and runs a bash script*

  _Given_ a developer has set up the sql tables from the zip file
  _When_ he/ she runs the script
  _Then_ missing columns are added to the db


pjotrp:
So now you can use the DB locally, right. Thanks!

PR #411, opened 2020-07-25
Updating docs

pjotrp:
@zsloan I have updated the install docs and added text on how to incorporate a locally installed python module (for development purposes). 

You may want to check my changing 

     'owner_id'    : "none", # webqtlConfig.DEFAULT_OWNER_ID,

since it was not defined in the source.


zsloan:
Owner ID can be changed in the resource manager (resource_manager.py function change_owner). That function is basically for adding a resource with default settings (including owner, which is Rob's user ID by default) when the resource doesn't already exist in Redis.

I think it's actually redundant now, though; there's a separate thing that checks if a user's ID is in super_users in Redis, and if so gives them owner-equivalent privileges. So I think it's okay to merge. I'll try some stuff on my branch first just to make sure, but if it's okay I'll merge in a few minutes.

zsloan:
Oh wait, I see what the confusion is. I hadn't committed the change adding that to webqtlConfig yet.

zsloan:
Do you think it might be a good idea to pass this in as an environment variable or something? I'm wondering about situations where someone is running their own GN2 branch and might want to set a different default owner, and I don't know if such people will feel comfortable changing super_users in Redis.

I'm wondering if I should commit the addition of DEFAULT_OWNER_ID to webqtlConfig or if I should just merge your changes (since it won't functionally make a difference for our GN2 instance).

pjotrp:
I think, for now, let's just have a default in DEFAULT_OWNER_ID. BTW webqtlConfig is kinda obsolete. We should set things in ./etc/default_settings.py

pjotrp:
@zsloan I'll merge. Just keep that parameter in mind.

PR #413, opened 2020-07-27
Minor fix for CTLscan, updated the parameter name: ncores to nthreads

DannyArends:
Parameter name was changed in the R package, updated in the GN2 source code


wqflask/wqflask/ctl/ctl_analysis.py
```diff
@@ -143,7 +143,7 @@ def run_analysis(self, requestform):
         #r_write_table(rPheno, "~/outputGN/pheno.csv")
 
         # Perform the CTL scan
-        res = self.r_CTLscan(rGeno, rPheno, strategy = strategy, nperm = nperm, parametric = parametric, ncores = 6)
+        res = self.r_CTLscan(rGeno, rPheno, strategy = strategy, nperm = nperm, parametric = parametric, nthreads = 6)
```
BonfaceKilz, line 146: **nitpick (non-blocking):** It would be helpful to indicate the exact version of the lib where the change happens. Also, is the CTLscan backward compatible? We manage libs using GUIX and I can see that we use python2-rpy2 v2.7.6. I'll run a migration to python3.8 soon so I'm thinking this would be applicable. Wdyt?

DannyArends: The change was introduced in the R package on CRAN when updated from 1.0.0-2 to 1.0.0-3
I think GN2/GUIX uses the latest version from CRAN, as such the parameter name change broke the Rpy2 wrapper code calling into the R package

BonfaceKilz (approved):

BonfaceKilz:
@DannyArends this was resolved in 8db56f247 during the Python3 migration.and in 18a6ada32ea8453a6ac939f20ef78b722d1b1b6c (when @zsloan worked on it). I'm going ahead to close this. Thanks for the contrib!

PR #415, opened 2020-07-29
Tests/add tests for gen menu

BonfaceKilz:
#### Description

This PR:
- Adds tests to **wqflask/wqflask/api/gen_menu.py**
- Cleans up **wqflask/wqflask/api/gen_menu.py**

### DIFF
--- a/wqflask/tests/api/test_gen_menu.py
+++ b/wqflask/tests/api/test_gen_menu.py
@@ -69,9 +69,11 @@ class TestGenMenu(unittest.TestCase):
 
     @mock.patch('wqflask.api.gen_menu.g')
     def test_get_species(self, db_mock):
-        """Test that assertion is raised when dataset and dataset_name are defined"""
-        db_mock.db.execute.return_value.fetchall.return_value = (('human', 'Human'),
-                                                                 ('mouse', 'Mouse'))
+        """Test that assertion is raised when dataset and dataset_name
+        are defined"""
+        db_mock.db.execute.return_value.fetchall.return_value = (
+            ('human', 'Human'),
+            ('mouse', 'Mouse'))
         self.assertEqual(get_species(),
                          [['human', 'Human'], ['mouse', 'Mouse']])
         db_mock.db.execute.assert_called_once_with(
@@ -84,7 +86,8 @@ class TestGenMenu(unittest.TestCase):
         db_mock.db.execute.return_value.fetchall.side_effect = [
             # Mouse
             (('BXD', 'BXD', None),
-             ('HLC', 'Liver: Normal Gene Expression with Genotypes (Merck)', 'Test')),
+             ('HLC', 'Liver: Normal Gene Expression with Genotypes (Merck)',
+              'Test')),
             # Human
             (('H_T1', "H_T", "DescriptionA"),
              ('H_T2', "H_T'", None))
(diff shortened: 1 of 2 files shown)


zsloan:
Ah, I'm guessing the SQL queries were switched to using single quotes in parenthesis in order to avoid the actual query being passed to MariaDB containing a bunch of spaces/newlines?

If so, I'll try to make similar changes to such queries as I encounter them.

BonfaceKilz:
Well, that depends. If you are sure nothing changes, then that'd be great. I only changed them to make it easier to test: https://github.com/genenetwork/genenetwork2/pull/415/commits/d50f72869b8d0a7078f0481aad2b04346cd56e3f#diff-e74417ee92f04fb38de592038107a99cR96-R98-- it's easier to test strings that don't have newline and extra spaces in them(the Python indentation spaces form part of the string because we used docstrings there). Notice how I check that the mocks are called with the correct constructed query. That way I'm sure I haven't changed the behaviour in any way. Also, for readability, I break the long string up into different lines, and concatenate them using the **+**

BonfaceKilz:
@zsloan could you review that ^^. Imho it's good to go. The only "logical" changes I've made are removing some unreachable **else** branch here: https://github.com/genenetwork/genenetwork2/pull/415/commits/b4a6a2cdc523d09334a33ca9a4efe7bc0e77d3de

zsloan:
I'm getting the following when I run test_gen_menu.py (which seems weird
since the same thing seems to work in other tests):
>
> Traceback (most recent call last):
>   File
> "/export/local/home/zas1024/gn2-zach/gene/wqflask/tests/api/test_gen_menu.py",
> line 16, in <module>
>     class TestGenMenu(unittest.TestCase):
>   File
> "/export/local/home/zas1024/gn2-zach/gene/wqflask/tests/api/test_gen_menu.py",
> line 70, in TestGenMenu
>     @mock.patch('wqflask.api.gen_menu.g')
> AttributeError: 'module' object has no attribute 'patch'

zsloan:
Oh, maybe I need to use a different GUIX build? Maybe the one I'm using
doesn't have mock installed?

zsloan:
Ah, I'm actually getting the same import error from the other tests in
gene/wqflask/tests/base.

BonfaceKilz:
A quick way to see if you have mock installed is to run: `ls $GN2_PROFILE/lib/python2.7/site-packages | grep mock` where `$GN2_PROFILE` is the location of your guix profile. What's your output? Most likely it's because you are using a different guix build. I've just ran into the same problem on Penguin2 and all I did was update the my guix profile with latest package defs from guix-bioinformatics. HTH!

zsloan:
It wasn't installed on the profile previously used in the alias, but it is
installed on a more recent one. After updating the alias to use the more
recent profile it still gives the same error, though.

BonfaceKilz:
Well, let's try manually troubleshooting things. How about getting into a python REPL e.g.:

```
env GN2_PROFILE=/home/bonface/opt/genenetwork2 TMPDIR=/home/bonface/tmp SERVER_PORT=5004 GENENETWORK_FILES=/home/bonface/gnu/data/gn2_data/ /home/bonface/projects/genenetwork2/bin/genenetwork2 /home/bonface/projects/genenetwork2/etc/default_settings.py -c
```
and then doing an
```
import mock
```

What do you get?

zsloan:
I found the issue (after noticing I could import it but "mock.__file__" was
just pointing to a mock/__init__.pyc); there was a mock directory still in
wqflask with just the pyc for an elastic search test in it. It was
apparently looking in that path before the one in the guix profile. After
removing that things worked fine.

BonfaceKilz:
Great! @pjotrp or @zsloan, we should merge this if things are ok \m/\m/

PR #456, opened 2020-10-08
Some trait page width changes

zsloan:
#### Description
Rob mentioned the trait page description text growing too wide with certain screen sizes, so I allowed it to shrink with screen size and also applied this to correlation and mapping method descriptions

Open the trait page for a trait like the one below and look at the Summary and at the right sides of the Mapping and Correlation sections (when expanded)
https://genenetwork.org/show_trait?trait_id=ILMN_1090075&dataset=GSE15222_F_RI_0409

N/A

N/A

### DIFF
--- a/wqflask/wqflask/templates/show_trait.html
+++ b/wqflask/wqflask/templates/show_trait.html
@@ -37,7 +37,7 @@
         <input type="hidden" name="covariates" value="">
         <input type="hidden" name="transform" value="">
 
-        <div class="container" style="min-width: 1450px;">
+        <div class="container" style="min-width: 700px;">
             <div class="panel-group" id="accordion">
                 <div class="panel panel-default">
                     <div class="panel-heading" data-toggle="collapse" data-parent="#accordion" data-target="#collapseOne" aria-expanded="true">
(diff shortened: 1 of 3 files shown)


BonfaceKilz (commented):
**Nitpick:** It would be nice if we moved things to a separate style-sheet. That way we'd have fine-grained control over screen widths using the @media property.

zsloan:
Yeah, this is definitely something that needs to be done at some point (not
just for this page, but for many others as well). Tomorrow or the next day
I might try and compile all of these style changes into a sheet for the
trait page.

BonfaceKilz:

> Yeah, this is definitely something that needs to
> be done at some point (not
> just for this page, but for many others as well).
> Tomorrow or the next day
> I might try and compile all of these style
> changes into a sheet for the
> trait page.
>

Great! We could track this on Trello or on GH then
work on it later \m/\m/.


-- 
Bonface M. K. (https://www.bonfacemunyoki.com)
Chief Emacs Mchochezi / Twitter: @BonfaceKilz
GPG key = D4F09EB110177E03C28E2FE1F5BBAE1E0392253F

zsloan:
Going to go ahead and merge this and just create a new PR for moving both these style changes and a lot of other styling to CSS files

PR #461, opened 2020-10-21
Mapping y axis fix

zsloan:
#### Description
Currently it's possible for the top Y axis tick in the mapping figure to be above the figure itself, causing it to sometimes intersect with the legend. This fix aims to prevent that.

For now, I'm just removing the top tick when it's higher than the top of the figure. It would probably be preferable to scale everything so that the top of the figure corresponds to the top tick, but that's a bit more involved (since all the different tracks need to be adjusted accordingly).

Try mapping the following trait using R/qtl (either with or without permutations) - https://genenetwork.org/show_trait?trait_id=10015&dataset=HXBBXHPublish
Previously this figure had a tick mark above the top of the figure. With the fix, this tick mark shouldn't be displayed.

N/A

N/A

### DIFF
--- a/wqflask/wqflask/marker_regression/display_mapping_results.py
+++ b/wqflask/wqflask/marker_regression/display_mapping_results.py
@@ -2074,9 +2074,7 @@ class DisplayMappingResults(object):
                         pass
 
             if self.permChecked and self.nperm > 0 and not self.multipleInterval:
-                if self.significant > LRS_LOD_Max:
-                    LRS_LOD_Max = self.significant * 1.1
-                #LRS_LOD_Max = max(self.significant, LRS_LOD_Max)
+                LRS_LOD_Max = max(self.significant, LRS_LOD_Max)
             else:
                 LRS_LOD_Max = 1.15*LRS_LOD_Max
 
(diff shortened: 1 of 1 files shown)


BonfaceKilz:
@zsloan I've run the tests locally and things work fine--- so nothing wrong with your code. Sorry for the false negative.

This is due to degraded performance in github actions. See: https://www.githubstatus.com/incidents/hvh1bq3ywhyp?utm_ts=1603355581 and https://twitter.com/githubstatus/status/1319195040959397888?s=20

@zsloan, @robwwilliams and @pjotrp I've been able to reproduce that by re-running a previously successful test(https://github.com/genenetwork/genenetwork2/runs/1291565822) which now fails. Big unexpected ouch! for us :disappointed:. Regardless, if it becomes a big pain-point, we can always move to Travis or CircleCI since we have a working docker container :smiley:

pjotrp:
> @zsloan, @robwwilliams and @pjotrp I've been able to reproduce that by re-running a previously successful test(https://github.com/genenetwork/genenetwork2/runs/1291565822) which now fails. Big unexpected ouch! for us disappointed. Regardless, if it becomes a big pain-point, we can always move to Travis or CircleCI since we have a working docker container

Since this is a Guix container we could also consider running them on our server if we can figure out a github commit hook. Maybe an interesting target for sheepdog if we can generate the log online.

BonfaceKilz:
The failing build only affects the python2 branch. I've re-ran things on the python3 branch and the tests pass. I've raised that on the GH community forum here: https://github.community/t/failing-tests-after-github-actions-degradation-resolution/139248. I won't spend too much time troubleshooting this since we'll move to python3 anyways...

PR #471, opened 2020-10-28
Bug/Replace map on strip with a list comprehension that calls strip()

BonfaceKilz:
#### Description

This PR replaces maps that call string.strip() with a list comprehension that calls strip on each element.

It fixes this error:
Screen Shot 2020-10-28 at 12 02 10 PM

- Reload the page from that :point_up:

You can't do `list(map(string.strip, header))` in Python3.

N/A

N/A

#### Questions

N/A

### DIFF
--- a/scripts/maintenance/readProbeSetSE_v7.py
+++ b/scripts/maintenance/readProbeSetSE_v7.py
@@ -72,14 +72,14 @@ GeneList = []
 isCont = 1
 header = fp.readline()
 header = header.strip().split('\t')
-header = list(map(string.strip, header))
+header = [item.strip() for item in header]
 nfield = len(header)
 line = fp.readline()
 
 kj = 0
 while line:
     line2 = line.strip().split('\t')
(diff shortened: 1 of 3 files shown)


zsloan (approved):
Looks good to me. I should probably look around to see if there are any other places that define lists like this instead of using list comprehensions.

BonfaceKilz:

> @zsloan approved this pull request.
>
> Looks good to me. I should probably look around to
> see if there are any other places that define
> lists like this instead of using list
> comprehensions.
Yeah. I kind of already did that by doing:

`ag "map\(string\.strip" -Gpy$`

FYI, ag is silver searcher
-- 
Bonface M. K. <https://www.bonfacemunyoki.com>
Chief Emacs Bazu / Rieng ya software sare
Mchochezi of: <https://upbookclub.com> / Twitter: @BonfaceKilz
GPG Key: D4F09EB110177E03C28E2FE1F5BBAE1E0392253F

PR #482, opened 2020-11-05
Feature/corr table changes

zsloan:
#### Description
This PR is intended to implement Scroller with the correlation page table and convert that correlation page table to using JSON input instead of drawing all the HTML and then converting that to DataTables.

It will also change the way the show/hide column buttons work to avoid the glossary link question marks being included in them.

Do correlations from any trait page

N/A

N/A


wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -484,30 +484,31 @@ def generate_corr_json(corr_results, this_trait, dataset, target_dataset, for_ap
...
         if target_dataset.type == "ProbeSet":
             results_dict['symbol'] = trait.symbol
+            if len(trait.description_display) > 40:
+                results_dict['description'] = trait.description_display[:40] + "..."
+            else:
+                results_dict['description'] = trait.description_display
             results_dict['description'] = trait.description_display
             results_dict['location'] = trait.location_repr
-            results_dict['mean'] = float(trait.mean)
+            if trait.mean and trait.mean != "":
```
BonfaceKilz: **suggestion:nitpick:** You could cast that into a bool:

``` python
#+begin_src python :results output
print(bool(""))

print(bool(None))

print(bool("some value"))

print(bool(2))
#+end_src

#+RESULTS:
: False
: False
: True
: True
```

So this should work:
```suggestion
            if bool(trait.mean) 
```

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -484,30 +484,31 @@ def generate_corr_json(corr_results, this_trait, dataset, target_dataset, for_ap
...
+                results_dict['mean'] = float(trait.mean)
+            else:
+                results_dict['mean'] = "N/A"
             if trait.LRS_score_repr != "N/A":
                 results_dict['lrs_score'] = "%.1f" % float(trait.LRS_score_repr)
             else:
                 results_dict['lrs_score'] = "N/A"
             results_dict['lrs_location'] = trait.LRS_location_repr
-            if trait.additive != "":
+            if trait.additive and trait.additive != "":
```
BonfaceKilz: See above

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -484,30 +484,31 @@ def generate_corr_json(corr_results, this_trait, dataset, target_dataset, for_ap
...
                 results_dict['additive'] = "%0.3f" % float(trait.additive)
             else:
                 results_dict['additive'] = "N/A"
-            if for_api:
-                results_dict['sample_r'] = "%0.3f" % float(trait.sample_r)
-            else:
-                results_dict['sample_r'] = "<a target='_blank' href='corr_scatter_plot?dataset_1=" + str(dataset.name) + "&dataset_2=" + str(trait.dataset.name) + "&trait_1=" + str(this_trait.name) + "&trait_2=" + str(trait.name) + "'>" + "%0.3f" % float(trait.sample_r) + "</a>"
+            results_dict['sample_r'] = "%0.3f" % float(trait.sample_r)
```
BonfaceKilz: **suggestion:nitpick:** Since we now use Python3, it's a nice idea to switch to [f-strings](http://zetcode.com/python/fstring):

```python
#+begin_src python :results output
float_string = "0.35535115"
print("%0.3f" % float(float_string))

print(f"{float(float_string):.3f}")
#+end_src

#+RESULTS:
: 0.355
: 0.355
```

So this would be:
```suggestion
            results_dict['sample_r'] = f"{float(trait.sample_r):.3f}"
```
Idea is to not do this *at* once. Instead, when we encounter a %-string(or when you need formatted strings), we simply default to f-strings.
cc @Alexanderlacuna

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -516,41 +517,49 @@ def generate_corr_json(corr_results, this_trait, dataset, target_dataset, for_ap
...
                     results_dict['pubmed_id'] = "N/A"
                     results_dict['year'] = "N/A"
                 else:
-                    results_dict['pubmed'] = "N/A"
+                    results_dict['pubmed_link'] = "N/A"
+                    results_dict['pubmed_text'] = "N/A"
             results_dict['lrs_score'] = trait.LRS_score_repr
             results_dict['lrs_location'] = trait.LRS_location_repr
-            if trait.additive != "":
+            if trait.additive and trait.additive != "":
```
BonfaceKilz: See previous comments...

wqflask/wqflask/static/new/css/trait_list.css
```diff
@@ -0,0 +1,34 @@
...
+    right: 10px;
+    background: rgba(0, 0, 0, 0.8);
+    color: white;
+    box-shadow: 3px 3px 10px rgba(0, 0, 0, 0.5);
+    text-align: right;
+    border-radius: 3px;
+    padding: 0.4em;
+    z-index: 2;
+    display: none;
+} */
```
BonfaceKilz: **suggestion:** Remove commented out blocks at the very end of this PR

Commented out blocks usually wind up getting stale and sometimes make it confusing to read the code...

wqflask/wqflask/static/new/css/trait_list.css
```diff
@@ -0,0 +1,34 @@
...
+
+table.dataTable tbody td {
+    /* word-wrap: break-word;
+    word-break: break-all; */
```
BonfaceKilz: See above.

BonfaceKilz (changes requested):

zsloan:
Thanks for the suggestions! I didn't think of using boolean casting that
way and forgot about the f-string syntax.

Alexanderlacuna:
@BonfaceKilz Thanks for the suggestions.

zsloan:
I think this should be ready now? I'll do separate PRs for converting the other tables to using Scroller in the same way.

PR #483, opened 2020-11-05
Refactor branch

Alexanderlacuna:
#### Description
Refactored natural_sort to return new list rather that modifying original

N/A


Alexanderlacuna:
Duplicate

PR #486, opened 2020-11-16
Initialize flask-socketio

BonfaceKilz:
#### Description

TODO

TODO

TODO

TODO

TODO

#### Questions

TODO

### DIFF
--- a/wqflask/run_gunicorn.py
+++ b/wqflask/run_gunicorn.py
@@ -9,14 +9,18 @@
 
 print("===> Starting up Gunicorn process")
 
+from flask_socketio import SocketIO
+
 from wqflask import app
 from utility.startup_config import app_config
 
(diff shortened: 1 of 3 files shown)


BonfaceKilz:
Been playing around with sockets here: https://github.com/BonfaceKilz/flasksocketio-playground/blob/main/run.py. Will try to wire that up in gn2. Was able to get stdout(for now reading line by line instead of char by char) and stream it to the flask template. Other things I tried were streaming flask(see: https://flask.palletsprojects.com/en/1.1.x/patterns/streaming/). Another thing I learnt is that you could use something like Redis or task managers like Celery to source for events(which would come in handy for Sheepdog). Challenges were parsing the terminal special chars for things like the guix command(I used that during testing to simulate running processes).

PR #491, opened 2020-11-25
Feature/manhattan plot coloring

zsloan:
#### Description
This adds the following options to manhattan plots:
- Giving all points a single color that can be selected by the user
- Make all chromosomes have different colored points (using preset colors)
- Use an alternating Red/Blue color scheme (this was previously the default)

The colors used in the "different colors for all chromosomes" option are subject to change (I'm color-blind I just chose a random list of contrasting colors I found online), and I also might want to add the option to select colors for the "Alternating" color scheme.

Do GEMMA mapping from any trait page (like the one below) and use the "Manhattan Plot Color Scheme" option in the mapping options area at the top of the page.

https://genenetwork.org/show_trait?trait_id=1436869_at&dataset=HC_M2_0606_P

N/A

N/A

### DIFF
--- a/wqflask/wqflask/marker_regression/display_mapping_results.py
+++ b/wqflask/wqflask/marker_regression/display_mapping_results.py
@@ -74,6 +74,34 @@ DARKVIOLET = ImageColor.getrgb("darkviolet")
 MEDIUMPURPLE = ImageColor.getrgb("mediumpurple")
 # ---- END: Define common colours ---- #
 
+# ZS: List of distinct colors for manhattan plot if user selects "varied"
+DISTINCT_COLOR_LIST = [
+    ImageColor.getrgb("#FF0000"),
+    ImageColor.getrgb("#00FF00"),
+    ImageColor.getrgb("#0000FF"),
+    ImageColor.getrgb("#FFFF00"),
+    ImageColor.getrgb("#FF00FF"),
+    ImageColor.getrgb("#00FFFF"),
+    ImageColor.getrgb("#000000"),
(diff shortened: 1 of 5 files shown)


wqflask/wqflask/marker_regression/display_mapping_results.py
```diff
@@ -74,6 +74,34 @@
...
+    ImageColor.getrgb("#00C000"),
+    ImageColor.getrgb("#0000C0"),
+    ImageColor.getrgb("#C0C000"),
+    ImageColor.getrgb("#C000C0"),
+    ImageColor.getrgb("#00C0C0"),
+    ImageColor.getrgb("#C0C0C0"),
+    ImageColor.getrgb("#400000"),
+    ImageColor.getrgb("#004000"),
+    ImageColor.getrgb("#000040"),
+]
```
BonfaceKilz: **nitpick:** Whenever I find myself doing something like this, I'd use a list comprehension to isolate the list in one place. Something like:

```
COLOR_LIST = ["A", "B" ...]
DISTINCT_COLOR_LIST = [ImageColor.getrgb(color) for color in COLOR_LIST]
```

BonfaceKilz (approved):
This LGTM although I don't know enough Biology(yet) to test it out(though I tried)...

wqflask/wqflask/marker_regression/display_mapping_results.py
zsloan: Ah, that's a good idea.

BonfaceKilz (approved):

PR #499, opened 2020-12-02
Feature/hide empty column

Alexanderlacuna:
#### Description
add script to hide column when values are N/A

import the script for any table

Screenshot from 2020-12-03 01-30-57
Screenshot from 2020-12-03 01-31-25

### DIFF
--- a/wqflask/wqflask/static/new/javascript/auto_hide_column.js
+++ b/wqflask/wqflask/static/new/javascript/auto_hide_column.js
@@ -1,22 +1,21 @@
+$(function(){
+	$("table").each(function(table){
+		$(this).find("th").each(function(col_index,col_th){
+			var filter_counter=0
+			var col_td=$(this).closest('table').find('tr td:nth-child(' + (col_index + 1) + ')')
+			col_td.each(function(td_index,col_td){
+				if(this.innerHTML==""||this.innerHTML=="N/A"){
+					filter_counter+=1
+				}
+				else{
+					return false
+				}
+			})
+			if (filter_counter==$(this).closest("table").find("tr").length-1){
+					$(this).hide();
+                   col_td.hide();
 
-$(function (){
-$("table th").each(function(th_index,col_th){
-	var filter_counter = 0;
-	// map each column to row ie  column i to ith element
-	var column_tds = $(this).closest('table').find('tr td:nth-child(' + (index + 1) + ')')
-	column_tds.each(function(td_index,col_td){
-		if(this.innerHTML==''||this.innerHTML=="N/A"){
-			filter_counter+=1;
-		}
-		else{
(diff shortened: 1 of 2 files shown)


zsloan:
A couple things I'm unsure about:
- I don't think this will work with Scroller, since all cells won't be present in the HTML; the table cell contents would need to be pulled with the DataTables API (all tables can assumed to be using DataTables). An example of getting table data this way is below*.
- I'm not sure how the speed will scale with a very large number of results (and how this would impact initial page load time). This was basically the biggest problem; I think I found some similar code that did this (though before using Scroller, so that same code wouldn't work now), but it was sluggish when doing a global search or something with many rows, since it would have to iterate through all the table's cells in the JS after the page loaded.

Using the DataTables API may actually help address the second point; it seems to be a lot faster than just iterating through elements using JQuery.

* https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/wqflask/static/new/javascript/show_trait.js Around line 615 is an example of this (among other places in that same file)

Alexanderlacuna:
@zsloan  Yeah you are right on scaling tried with 20,000 rows 4 columns, seems a bit sluggish. will try your approach and see how that goes.

zsloan:
I'll look at this tomorrow, forgot to look at it today (just to let you know that I saw it).

zsloan:
After testing this, I think that it might be treating the columns as directly corresponding with the data passed to DataTables, but most of the tables in question have several columns that are just rendered using the JSON `data.`

For example, starting around line 260 of this file ( https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/wqflask/templates/search_result_page.html ), the columns are defined, with some directly sourcing their data from the JSON, but others being rendered from multiple fields (like the Record column). So the number of fields in data doesn't directly correspond to the number of columns. 

I think that the core loop just needs to refer to the nodes instead of the data. I tried making a change that seems to work as long as deferRender isn't being used. I'm not sure how to submit the change to the existing PR, but the code block for the loop (that in this case has the outer loop be columns and inner loop be rows) that I tried (and seems to work when I tested by adding an extra column of nothing but N/A values) is below:

```
    let validColumns = [];
    for (let col=0; col<columnCount; col++){
        colObj = datatable.column(col).nodes().to$();
        allNAs = true;
        for (let i=0;i<numberOfRows;i++){
            cellContent = colObj[i].childNodes[0].data
            if (cellContent != "N/A" && cellContent != ""){
                allNAs = false;
                break;
            }
        }
        if (!allNAs){
            validColumns.push(col)
        }
    }
```

Alexanderlacuna:
@zsloan  added your changes to this PR.

zsloan:
Seems fine to me!

PR #503, opened 2020-12-05
update  tests for run_gemma

Alexanderlacuna:
#### Description
modify tests for run_gemma

as unittest

N/A

N/A
#### Questions

N/A

### DIFF
--- a/wqflask/tests/unit/wqflask/marker_regression/test_gemma_mapping.py
+++ b/wqflask/tests/unit/wqflask/marker_regression/test_gemma_mapping.py
@@ -53,12 +53,12 @@ class TestGemmaMapping(unittest.TestCase):
         this_chromosomes={}
         for i in range(1, 5):
             this_chromosomes[f'CH{i}']=(AttributeSetter({"name": f"CH{i}"}))
-        chromo = AttributeSetter({"chromosomes": this_chromosomes})
+        chromosomes = AttributeSetter({"chromosomes": this_chromosomes})
 
         dataset_group = MockGroup(
             {"name": "GP1", "genofile": "file_geno"})
         dataset = AttributeSetter({"group": dataset_group, "name": "dataset1_name",
-                                   "species": AttributeSetter({"chromosomes": chromo})})
+                                   "species": AttributeSetter({"chromosomes": chromosomes})})
         trait = AttributeSetter({"name": "trait1"})
         samples = []
         mock_gen_pheno_txt.return_value = None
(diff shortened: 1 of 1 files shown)


BonfaceKilz (approved):
Much thanks for this. Actually ran into these failing tests in my branch and this seems to fix things

zsloan:
Ah, I think this might have been a similar issue to the recent one I noticed with gemma_mapping.py (I had to change the way this_chromosomes was set).

BonfaceKilz:

> Ah, I think this might have been a similar issue
> to the recent one I noticed with gemma_mapping.py
> (I had to change the way this_chromosomes was
> set).

Yup and on point... I could see that bug was
introduced here:
7aaa4b87fcf405c32c11e2e2d664941c43c14430

-- 
Bonface M. K. <https://www.bonfacemunyoki.com>
Humble GNU Emacs User / Bearer of scheme-y parens
Curator: <https://upbookclub.com> / Twitter: @BonfaceKilz
GPG Key: D4F09EB110177E03C28E2FE1F5BBAE1E0392253F

pjotrp:
Is this a first with the new test system?

BonfaceKilz:
Pjotr Prins <notifications@github.com> writes:

> Is this a first with the new test system?

Yup! I try to pay keen attention when sth is
refactored, but this one slipped my eye(probably
because that list comprehension made sense); but
was caught by @zsloan.

-- 
Bonface M. K. <https://www.bonfacemunyoki.com>
Humble GNU Emacs User / Bearer of scheme-y parens
Curator: <https://upbookclub.com> / Twitter: @BonfaceKilz
GPG Key: D4F09EB110177E03C28E2FE1F5BBAE1E0392253F

zsloan:
Yeah, I think the confusion was related to the Species class being really
weird. Species has a "chromosomes" attribute that is of the "Chromosomes"
class, but the "Chromosomes" class also has its own attribute which is also
called "chromosomes."

I feel like there isn't really any need for there to be a "Chromosomes"
class here, and that it could just be replaced by a function that returns
the list of chromosomes using either species or dataset as input.

BonfaceKilz:

> Yeah, I think the confusion was related to the Species class being really
> weird. Species has a "chromosomes" attribute that is of the "Chromosomes"
> class, but the "Chromosomes" class also has its own attribute which is also
> called "chromosomes."
>

Oof!

> I feel like there isn't really any need for there to be a "Chromosomes"
> class here, and that it could just be replaced by a function that returns
> the list of chromosomes using either species or dataset as input.
>

With time we'll clean those up. I optimistically
think we have a good momentum and we'll soon get
there IMHO :)

-- 
Bonface M. K. <https://www.bonfacemunyoki.com>
Humble GNU Emacs User / Bearer of scheme-y parens
Curator: <https://upbookclub.com> / Twitter: @BonfaceKilz
GPG Key: D4F09EB110177E03C28E2FE1F5BBAE1E0392253F

PR #514, opened 2020-12-08
Replace colorbox lib

Alexanderlacuna:
#### Description
replaced jquery.colorbox.js and replaced /packages/colorbox/example4/colorbox.css with new/css/colorbox.css


BonfaceKilz (approved):
This LGTM. On Guix, we have jQuery Colorbox v1.4.36 and on static/packages dir, we had v1.4.15. That's the only difference AFAICT. Since they are the same major version, there should be no discrepancies.

PR #515, opened 2020-12-08
Replace datatable libs

Alexanderlacuna:
#### Description
(scroller.dataTables,buttons.dataTables,datatable jquery)


BonfaceKilz (approved):
This LGTM! Only comment (more to @zsloan) is that the dir: /static/new/packages/DataTables --- by the time of this merging was removed in d2bb95c5e169f1a616ebbea75bbe15b788b0fe70. This wasn't caught; but this PR makes some of those files available.

PR #516, opened 2020-12-08
Replace jscolor lib

Alexanderlacuna:
#### Description
replace jscolor lib


BonfaceKilz (approved):
The version in Guix and the one being replaced match exactly.

PR #517, opened 2020-12-08
Replace jszip lib

Alexanderlacuna:
#### Description
replace jszip lib


BonfaceKilz (approved):
Jszip in guix is v3.2.1 and I'm not entirely sure what version the one which we are replacing is. I don't think this will bring problems though.

PR #522, opened 2020-12-08
Feature/default scatterplot tab

zsloan:
#### Description
This PR makes any scatter-plot links from a Spearman correlation default to the Spearman scatter-plot. Previously it would just default to Pearson while having a tab for both methods.

Open a scatterplot from either the Correlation Matrix or Correlation Results pages

N/A

N/A


BonfaceKilz (approved):
This LGTM

PR #523, opened 2020-12-09
Replace images

Alexanderlacuna:
#### Description
replaces images used in css


BonfaceKilz (approved):
I would have surely missed these! Thanks for catching these since I grepped only html files(and never thought of the css files)!

PR #525, opened 2020-12-09
Replace (underscore string,underscore js)

Alexanderlacuna:
#### Description
replaces underscore string and underscore  js


BonfaceKilz (approved):
AFAICT the version we have here and what we have in Guix are different. In guix we have: 1.9.1; Not sure what version we use in gn2 though. @zsloan could you help out with that. I'll go ahead and merge this.

PS: FWIW, we can choose to use the latest version of this upstream. @zsloan @Alexanderlacuna WDYT?

pjotrp:
Bring in latest. If it breaks we'll take another look.

BonfaceKilz:
Done!
http://git.genenetwork.org/guix-bioinformatics/guix-bioinformatics/commit/1eb8f35d65afd67cfd8653d06e39c1f4d2ff6588

PR #401, opened 2020-06-16
Feature/integrate feed analyzer

BonfaceKilz:
#### Description

This PR adds code to read code from the REDIS queues that contain the scored tweets. For now, it display 15 tweets; we could always tweak it to read more

#### How should this be tested

The test server will be shared later in the comments section

#### Any background context you want to provide

N/A

#### What are the relevant pivotal tracker stories

https://github.com/genenetwork/genenetwork2/issues/400

#### Screenshots (if appropriate)

N/A

#### Questions

N/A

PR #526, opened 2020-12-09
Replace twitter-post-fetcher

Alexanderlacuna:


BonfaceKilz (approved):
We  don't really use this library(it should be replaced  with the new feed). Related PR: https://github.com/genenetwork/genenetwork2/pull/401. Any way, going ahead to merge this.

PR #528, opened 2020-12-09
Replace shapiro wilk lib

Alexanderlacuna:


BonfaceKilz (approved):
There's a mismatch with what we have in Guix, but both libs appear to be doing the same thing...

PR #529, opened 2020-12-10
Feature/add mapping endpoints

Alexanderlacuna:
#### Description
update /run_mapping and /loading in wqflask/views.py to have each mapping on  own path

### DIFF
--- a/wqflask/wqflask/views.py
+++ b/wqflask/wqflask/views.py
@@ -657,135 +657,26 @@ def loading_page():
 
     return rendered_template
 
-@app.route("/run_mapping", methods=('POST',))
-def mapping_results_page():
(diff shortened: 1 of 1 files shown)


zsloan:
We actually don't need an endpoint for Plink, by the way (it was basically made obsolete by GEMMA).

I'm also making some changes that will probably be relevant to the get_loading_page_data function (and maybe some other places) here - https://github.com/genenetwork/genenetwork2/pull/535

Currently sample data is passed as a bunch of separate inputs (where it says "if key in wanted or **key.startswith(('value:'))**" in the code - currently there are separate inputs with names like "value:BXD1", "value:BXD2", etc), and this PR changes the sample values to be passed as JSON instead.

PR #531, opened 2020-12-14
add test  for snp browser and utility

Alexanderlacuna:
#### Description
adds tests for snp browser and utility

as unittest


BonfaceKilz (commented):
Ping @Alexanderlacuna it's a good idea to start incrementally getting rid of some of the type checks since those are in-builts in Python3. Wdyt?

Alexanderlacuna:
@BonfaceKilz  Cool forgot about that.

BonfaceKilz (approved):
LGTM

PR #541, opened 2021-01-12
Feature/swap legend metadata positions

zsloan:
#### Description
This PR swaps the position of metadata and the legend in the mapping results figure.

I think this is mainly because metadata can be wider (so it has more space to expand if it's on the left) and is generally more important, and legends are more typically on the right.

Do mapping from any trait page and look at the figure to see if it looks okay.

N/A

N/A


zsloan:
I think the only thing remaining is to fix the position of the "Sequence Site" legend.

PR #548, opened 2021-01-29
Feature/add upload api

BonfaceKilz:
#### Description

Add endpoint for adding gzipped endpoint. To this enpoint you pass in a tar.gz file that contains the metadata and phenotype file.

`curl -F file=@upload-data.tar.gz http://localhost:5004/api/v_pre1/metadata/upload`

Here's the spec: https://github.com/genenetwork/gn-docs/blob/master/api/upload.md

On Trello

N/A

#### Questions

N/A


pjotrp (commented):
Can we use tmpdir instead of hard coded?

pjotrp (commented):
What type of tests do we need to put in place?

BonfaceKilz:
Pjotr Prins <notifications@github.com> writes:

> @pjotrp commented on this pull request.
>
> Can we use tmpdir instead of hard coded?

Yup!
-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Humble GNU Emacs User / Bearer of scheme-y parens
Curator: <https://upbookclub.com> / Twitter: @BonfaceKilz

BonfaceKilz:
Pjotr Prins <notifications@github.com> writes:

> @pjotrp commented on this pull request.
>
> What type of tests do we need to put in place?

Unit tests are easiest for this. I'll break up the
side-effectful function and write tests for the
broken pieces; and then finally for the actual
end-point add an integration test where you mock
the request. Not hard to do IMHO; I've done that
before.

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Humble GNU Emacs User / Bearer of scheme-y parens
Curator: <https://upbookclub.com> / Twitter: @BonfaceKilz

pjotrp:
Great. Let's add simple tests from the start. BTW I like the code. No hint of self :)

zsloan:
This probably isn't important in terms of functionality, but it might make sense to set the directory using TEMPDIR from this file (just because that's where most other files get the temp directory, like gemma_mapping.py, etc) - https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/utility/tools.py

pjotrp:
> This probably isn't important in terms of functionality, but it might make sense to set the directory using TEMPDIR from this file (just because that's where most other files get the temp directory, like gemma_mapping.py, etc) - https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/utility/tools.py

Yes, all global settings are handled through tools.py. We should rename that file ;)

BonfaceKilz:
Closing this since it's already added in gn3

PR #556, opened 2021-02-18
Bug/fix auth tests

zsloan:
#### Description
This fixes a few broken tests. I'm requesting a review on it just because it's the first time I've made changes to existing tests and it took me a while to figure out mocking/patching worked in test_authentication_tools.py, but the changes I made seem to cause all tests to pass.

All unit tests should pass now.

### DIFF
--- a/wqflask/tests/unit/utility/test_authentication_tools.py
+++ b/wqflask/tests/unit/utility/test_authentication_tools.py
@@ -5,7 +5,6 @@ from unittest import mock
 from utility.authentication_tools import check_resource_availability
 from utility.authentication_tools import add_new_resource
 
-
 class TestResponse:
     """Mock Test Response after a request"""
     @property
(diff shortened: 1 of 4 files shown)


zsloan:
The one failed test in test_run_mapping should work properly when merged, since I pushed the change that includes the missing transform/covariates/n_samples fields in export_mapping_results.

wqflask/tests/unit/utility/test_authentication_tools.py
```diff
@@ -38,14 +37,15 @@ class TestCheckResourceAvailability(unittest.TestCase):
     """Test methods related to checking the resource availability"""
     @mock.patch('utility.authentication_tools.add_new_resource')
     @mock.patch('utility.authentication_tools.Redis')
-    @mock.patch('utility.authentication_tools.g', mock.Mock())
+    @mock.patch('utility.authentication_tools.g', TestUserSession())
```
BonfaceKilz, line 40: **nitpick**: It's best to pass in mock objects as patches instead of passing actual objects since you can make assertions like checking what the method was passed with.

BonfaceKilz (approved):
@zsloan You are right! Thanks for the fix!

wqflask/tests/unit/utility/test_authentication_tools.py
zsloan: I changed this because it was throwing an error when using mock.Mock() at the line that says "str.encode(g.user_session.user_id)" in line 34 of utility/authentication_tools.py

Though now that I think about it, it might be possible to change that line to convert the contents of Redis.smembers("super_users") to strings instead of converting g.user_session.user_id to bytes, which should make the original line in the unit test work.

PR #565, opened 2021-04-27
Stopped using the scanone function pointer when doing R/qtl mapping, …

zsloan:
#### Description
rpy2 and numpy were updated as a result of covariates no longer working with R/qtl, but after updating them there was an issue where marker names could no longer be extracted from the results returned by scanone (the R/qtl mapping function).

It turns out that this is because scanone was being run using a pointer to the R function, which caused its results to automatically be converted to a Python numpy object. Something about updating the above libraries (rpy2/numpy) caused this object to no longer contain the marker names as row names.

As a fix, I changed rqtl_mapping.py to run scanone directly as R code and then extracted the marker names from that and assigned them to a Python variable.

This whole situation highlights the need to change the convoluted way we use rpy2 with R/qtl.

Do any R/qtl mapping with a covariate selected


wqflask/wqflask/marker_regression/rqtl_mapping.py
```diff
@@ -376,30 +384,30 @@ def process_pair_scan_results(result):
 
 def process_rqtl_perm_results(num_perm, results):
     perm_vals = []
-    for line in str(results).split("\n")[1:(num_perm+1)]:
-        #print("R/qtl permutation line:", line.split())
-        perm_vals.append(float(line.split()[1]))
+    for item in results:
+        perm_vals.append(item)
 
     perm_output = perm_vals
```
BonfaceKilz: **Nitpick:** But a good place for a list comprehension :wink: 

```python
perm_vals = [item for item in results]
```

PR #577, opened 2021-05-13
minor fixes for correlation

Alexanderlacuna:
#### Description
 PR improves on things like u

1. Fixing ui issues for correlation template e.g displaying dataset_name and trait_name
2. integrating computing all correlation to the ui


BonfaceKilz (approved):
LGTM! We may have to re-work some of the things here; like the merge_correlation_results. I'll have a keener look at that once I'm done with the uploader work :)

PR #584, opened 2021-06-04
Feature/add scroller gsearch

zsloan:
#### Description
This implements Scroller for the global search pages + sets column widths and fixes a couple bugs.

This change undoes server-side processing, since it doesn't work well with Scroller as it is now. I think the problem is that the server-side processing code is written to load pages when using pagination, and I think with Scroller it's basically acting as if it's loading a single page consisting of all results every time Scroller needs to load in rows. It seems to work fine without server-side processing, though it might be relevant if we're loading over 10,000 rows.

Do either gene or phenotype global search and try scrolling.

### DIFF
--- a/wqflask/wqflask/gsearch.py
+++ b/wqflask/wqflask/gsearch.py
@@ -92,19 +92,30 @@ class GSearch:
                     if (line[8] != "NULL" and line[8] != "") and (line[9] != 0):
                         this_trait['location_repr'] = 'Chr%s: %.6f' % (
                             line[8], float(line[9]))
-                    try:
-                        this_trait['mean'] = '%.3f' % line[10]
-                    except:
-                        this_trait['mean'] = "N/A"
+
                     this_trait['LRS_score_repr'] = "N/A"
+                    this_trait['additive'] = "N/A"
+                    this_trait['mean'] = "N/A"
+
                     if line[11] != "" and line[11] != None:
-                        this_trait['LRS_score_repr'] = '%3.1f' % line[11]
+                        this_trait['LRS_score_repr'] = f"{line[11]:.3f}"
+                    if line[14] != "" and line[14] != None:
+                        this_trait['additive'] = f"{line[14]:.3f}"
+                    if line[10] != "" and line[10] != None:
+                        this_trait['mean'] = f"{line[10]:.3f}"
+
+                    locus_chr = line[16]
+                    locus_mb = line[17]
+
+                    max_lrs_text = "N/A"
+                    if locus_chr and locus_mb:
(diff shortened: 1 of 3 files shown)


wqflask/wqflask/gsearch.py
```diff
@@ -92,19 +92,30 @@ def __init__(self, kw):
...
+                    this_trait['additive'] = "N/A"
+                    this_trait['mean'] = "N/A"
+
                     if line[11] != "" and line[11] != None:
-                        this_trait['LRS_score_repr'] = '%3.1f' % line[11]
+                        this_trait['LRS_score_repr'] = f"{line[11]:.3f}"
+                    if line[14] != "" and line[14] != None:
+                        this_trait['additive'] = f"{line[14]:.3f}"
+                    if line[10] != "" and line[10] != None:
+                        this_trait['mean'] = f"{line[10]:.3f}"
```
BonfaceKilz, line 105: **suggestion**: This: `if lines[11] != None` can be expressed as: `if lines[11]`

BonfaceKilz:
@zsloan For CSS things, we should consider using more of [bootstrap](https://getbootstrap.com/docs/3.3/)

wqflask/wqflask/gsearch.py
zsloan: Oh yeah, forgot that empty strings are considered False for things like this also.

zsloan:
> @zsloan For CSS things, we should consider using more of [bootstrap](https://getbootstrap.com/docs/3.3/)

Are you talking about the div widths? I know how you can set widths in bootstrap using its grid system (with classes like col-xs-3, etc), but there are a number of situations like this where setting table width relative to window width doesn't work well (so in this case the min-width is meant to prevent the table width from shrinking beyond where it starts to look bad).

BonfaceKilz:

>> @zsloan For CSS things, we should consider using
> more of
> [bootstrap](https://getbootstrap.com/docs/3.3/)
>
> Are you talking about the div widths?

Yup. In addition to that, there's a styling for
tables provided by bootstrap and it's extensions.

> I know how you can set widths in bootstrap using
> its grid system (with classes like col-xs-3,
> etc), but there are a number of situations like
> this where setting table width relative to
> window width doesn't work well (so in this case
> the min-width is meant to prevent the table
> width from shrinking beyond where it starts to
> look bad).

Ah I see!

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist
Humble GNU Emacs User | Bearer of scheme-y parens
Curator: <https://upbookclub.com> | Twitter: @BonfaceKilz

PR #590, opened 2021-07-06
replace rpy2  for tissue correlation

Alexanderlacuna:
#### Description
this pr replaces rpy2 used to compute tissue correlation with
the function provided by gn3 to compute correlation.Also this pr fixes  the issue of broken links
in workflow


arunisaac:
Looks good to me, overall! But, could you split off the whitespace
changes (particularly those in
`wqflask/wqflask/correlation/show_corr_results.py` and in the
`get_trait_symbol_and_tissue_values` function of
`wqflask/wqflask/correlation/correlation_functions.py`) into a
separate commit? The whitespace changes distract from the main intent
of the commit. We want commits to be as atomic as possible.

Also, why is the mailing list link broken? I guess I'm missing some
background here. Maybe, you could fill me in.

Cheers!

Alexanderlacuna:
@arunisaac  Thanks for the review as part of genenetwork continuous integration we  check for links that aren't functioning in 
all gene network pages in order to improve on the user experience that is the reason test for broken links currently fails solution is to change to text until an alternative is found.

PR #605, opened 2021-09-25
Clustered heatmaps

fredmanglis:
#### Description

This pull request provides UI elements for clustered heatmaps, and also the Javascript code to send the requests and populate the page with the heatmap image.

In the `Collections` page, click on the **Clustered Heatmap** button after selecting at least 2 traits.

This provides some of the functionality [detailed here](https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/clustering.gmi)

https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/clustering.gmi

N/A

#### Questions

N/A


fredmanglis:
@BonfaceKilz Please have a look and see whether you can get this merged over the weekend so that Rob can start playing around with it from Monday.

zsloan:
I'm trying to test this and I'm having trouble with connecting to GN3. I don't know if it might have something to do with the connection being made using AJAX to localhost, but it says connection refused.

One other issue now that I'm seeing the interface is that it's probably best to have this function in the same way as the other collection features and have it open the figure in a new tab (rather than within the collections page itself). Opening figures within the collections page is actually pretty cool, but it's currently not really consistent with the way the other functions work.

I don't really mind making this change myself. You already did the actual meaningful work of returning the heatmap data, so I can just have it open the heatmap in its own template (a heatmap template already exists; yours would probably just be replacing it) and get the heatmap JSON through a server-side request that is then just passed to the template (since I know the request to GN3 through localhost works fine when doing in the Python code server-side).

fredmanglis:
@zsloan 

> I'm trying to test this and I'm having trouble with connecting to GN3. I don't know if it might have something to do with the connection being made using AJAX to localhost, but it says connection refused.
> 

Well the system uses the configuration variable `GN_SERVER_URL` in *genenetwork2/etc/default_settings.py* to find out where the GN3 server is. You will need to set that variable up correctly to point to the GN3 server, so that it can get the data from the GN3 server as appropriate.

> One other issue now that I'm seeing the interface is that it's probably best to have this function in the same way as the other collection features and have it open the figure in a new tab (rather than within the collections page itself). Opening figures within the collections page is actually pretty cool, but it's currently not really consistent with the way the other functions work.
> 

Well, maybe. I had shown the proof-of-concept to @pjotrp , and after a discussion, it was agreed to leave it as is, so that Rob can have a look, and get back to us with any issues.

> I don't really mind making this change myself. You already did the actual meaningful work of returning the heatmap data, so I can just have it open the heatmap in its own template (a heatmap template already exists; yours would probably just be replacing it) and get the heatmap JSON through a server-side request that is then just passed to the template (since I know the request to GN3 through localhost works fine when doing in the Python code server-side).

Please don't do that for now, until we have some feedback from Rob, especially on whether or not the heatmap satisfies the requirements.

pjotrp:
Yes, I like the inline version.

fredmanglis:
@zsloan 

I've added some changes to this, which should hopefully provide the user with the appropriate UI elements to allow for vertical/horizontal orientation of the heatmaps.

I have not been able to test out the latest changes, however, (having ssh issues on my end), so feel free to fix whatever minor issues come  up.

zsloan:
Thanks; I'll take a look today.

fredmanglis:
@zsloan Please have a look at this, and merge it. I did some minor updates to the code and added some documentation.

etc/default_settings.py
```diff
@@ -25,7 +25,6 @@
 import sys
 
 GN_VERSION = open("../etc/VERSION", "r").read()
-GN_SERVER_URL = "http://localhost:8880/" # REST API server
```
BonfaceKilz, line 28: **question:** How is this configured?

BonfaceKilz: Ah, I see it's in utils.

BonfaceKilz (approved):
This LGTM. +1 on adding the docs.

PR #611, opened 2021-10-15
Refactor/integrate gn proxy auth mechanism

BonfaceKilz:
#### Description

This PR:

- Deletes some unused logger imports
- Moves markdown_routes to api/markdown. This is more semantic.
- Adds a default setting to our configs for REDIS
- Deletes dead code in redis_utils
- Makes GN2_PROXY a configurable option
- Removes old routes for managing resources
- Adds an OrderedEnum class that's used to represent roles.
- Add extra decorators:
  - `@login_required`
  -  `@edit_access_required`
- Adds the following endpoints:
  - GET /resource-management/resources/<resource_id>
  - POST /resource-management/resources/<resource-id>/make-public
  - POST /resources/<resource_id>/change-owner
 
At the moment, this can't be tested. The user's hash and resources column need to be looked at again before we use this!

See: https://github.com/genenetwork/gn-gemtext-threads/blob/main/issues/authorisation.gmi

https://github.com/genenetwork/gn-gemtext-threads/blob/main/issues/authorisation.gmi

image

image

#### Questions

N/A


zsloan:
One minor thing is that I think the default mask would actually be something like view/view/not-admin (for data/metadata/admin respectively). Most of the time data will be public, so the default is for users to be able to view it (but obviously not edit it).

In line 105 how exactly is the "max(v)" working? Is that making use of the `__lt__` in OrderedEnum to compare indices of each access level?

It's interesting and helpful seeing the way you do something like this. It's much better than before, but in a way that is easy to understand and often makes me think "why didn't I do things like that?"

BonfaceKilz:

> One minor thing is that I think the default mask
> would actually be something like
> view/view/not-admin (for data/metadata/admin
> respectively).

Nice observation there:

```
    access_role = {
        "data": [DataRole.NO_ACCESS],
        "metadata": [DataRole.NO_ACCESS],
        "admin": [AdminRole.NOT_ADMIN],
    }
```

I use the lowest(!) access-role in the data there
is. That's because, as is, every resource has a
default mask which sets the access-role to "VIEW"
anyways. In other words, I'm assuming that in the
case of no "default-mask", the data should not be
accessible. Which is a good IMHO assumption to
make.

> Most of the time data will be
> public, so the default is for users to be able to
> view it (but obviously not edit it).

See above.

> In line 105 how exactly is the "max(v)" working?
> Is that making use of the __lt__ in OrderedEnum to
> compare indices of each access level?

Ah that. I prefer immutable types, hence I use the
`Enum` class. However, by default, the `Enum` type
does not have any way of comparing it's elements
unless they are integers. So for that I use the
OrderedEnum. As an example, with `OrderedEnum`
with this example:

```
@unique
class DataRole(OrderedEnum):
    NO_ACCESS = "no-access"
    VIEW = "view"
    EDIT = "edit"
```

`max(Datarole.NO_ACCESS, Datarole.VIEW,
Datarole.EDIT)`, will yield "Datorole.NO_ACCESS"
with it's string repr being "no-access".

> It's interesting and helpful seeing the way you do
> something like this. It's much better than before,
> but in a way that is easy to understand and often
> makes me think "why didn't I do things like that?"

You get better with time ;) ;)

-- 
Bonface M. K.
D4F09EB110177E03C28E2FE1F5BBAE1E0392253F (hkp://keys.gnupg.net)
Free Software Activist
Humble GNU Emacs User | Bearer of scheme-y parens
Curator: <https://upbookclub.com> | Twitter: @BonfaceKilz

BonfaceKilz:
@zsloan let's avoid using utility.tools to define configuration options; and instead use default_settings.py because of the following circular imports issues:

```
Traceback (most recent call last):
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/arbiter.py", line 589, in spawn_worker
    worker.init_process()
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/workers/base.py", line 134, in init_process
    self.load_wsgi()
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/workers/base.py", line 146, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/app/wsgiapp.py", line 58, in load
    return self.load_wsgiapp()
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/app/wsgiapp.py", line 48, in load_wsgiapp
    return util.import_app(self.app_uri)
  File "/gnu/store/286imikhis0dvb5aqcy87q6x8542xpw9-gunicorn-20.1.0/lib/python3.8/site-packages/gunicorn/util.py", line 359, in import_app
    mod = importlib.import_module(module)
  File "/gnu/store/sd2ic7bpv8fx3imy1j8xcjclx71sv6q0-python-3.8.2/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/bonface/projects/genenetwork2/wqflask/wsgi.py", line 1, in <module>
    from run_gunicorn import app as application  # expect application as a name
  File "/home/bonface/projects/genenetwork2/wqflask/run_gunicorn.py", line 12, in <module>
    from wqflask import app
  File "/home/bonface/projects/genenetwork2/wqflask/wqflask/__init__.py", line 12, in <module>
    from wqflask.resource_manager import resource_management
  File "/home/bonface/projects/genenetwork2/wqflask/wqflask/resource_manager.py", line 19, in <module>
    from wqflask.decorators import edit_access_required
  File "/home/bonface/projects/genenetwork2/wqflask/wqflask/decorators.py", line 12, in <module>
    from utility.tools import GN_PROXY_URL
  File "/home/bonface/projects/genenetwork2/wqflask/utility/tools.py", line 8, in <module>
    from wqflask import app
ImportError: cannot import name 'app' from partially initialized module 'wqflask' (most likely due to a circular import) (/home/bonface/projects/genenetwork2/wqflask/wqflask/__init__.py)
```

ATM, with this PR, there are 2 different places where the auth proxy are defined...

PR #631, opened 2021-11-16
Bug/fix correlation samples issue

zsloan:
#### Description
This PR is meant to fix a few current issues with correlations:

- The DB name is wrong in a query, causing correlations to throw an error
- After fixing the former, there are a couple other errors associated with metadata caching; this PR just makes the try/excepts not throw errors and doesn't fix the underlying issue
- Makes the sample_group (primary/other/all) selection actually work with correlations; it previously didn't

Test that correlations work as expected


zsloan:
I believe Alex dealt with the metadata caching issue himself, so this PR can be closed.

PR #632, opened 2021-11-16
Tempfix for reading from cache

Alexanderlacuna:
#### Description
tempfix for reading from cached data


Alexanderlacuna:
@zsloan  we are back to default. The read errors seem a bit odd I will have a look at things tomorrow.
plus with the  bugfix on  fetching samples, I will have to modify how we caching code

PR #643, opened 2021-12-09
Feature/warning messages for users

Alexanderlacuna:
Pr refactor code for gn3 wgcna:
makes these modification:
 - parse powers from the form data to compute softthreshold
 -  handle exception and generate error template if requests fails
 
*  To test select the powers to use in the form input and run the wgcna

### DIFF
--- a/wqflask/wqflask/templates/test_wgcna_results.html
+++ b/wqflask/wqflask/templates/test_wgcna_results.html
@@ -70,6 +70,10 @@
 }
 </style>
 <div class="container">
+    {% if  error!='null' %}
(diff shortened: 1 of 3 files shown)


wqflask/wqflask/templates/test_wgcna_results.html
```diff
@@ -70,6 +70,10 @@
 }
 </style>
 <div class="container">
+    {% if  error!='null' %}
```
BonfaceKilz, line 73: **nitpick:** Usually try to have a space around your operands.

BonfaceKilz (approved):
I see nothing wrong with this; so I will go ahead and have this merged.

PR #659, opened 2022-01-18
Fix some search page issues

zsloan:
This PR fixes a few issues:

- There was a bug where you could only add a limited number of rows to a collection at once (this is by far the most important issue). This is because the JS was still using JQuery to select rows, which doesn't select all rows if a table is using Scroller and has more than 150-200 rows (since Scroller doesn't load all rows in). I just changed the JS to use the DataTables API instead. There's still a remaining unrelated issue where loading the View Collection page is very slow.
- Fixes an issue where Genotype dataset queries were fetching from the Geno.chr_num column instead of Geno.Chr column. At least some datasets don't have any value set in the chr_num columns, leading some searches to return unknown chromosomes for the locations in their results.
- Fixed some aesthetic issues, most of which seemed to be unique to certain environments (so there might need to be further updates as I check with Rob to see if he still has issues). Most of these involve changing widths - the more detailed descriptions are in the commits.

### DIFF
--- a/wqflask/wqflask/static/new/javascript/search_results.js
+++ b/wqflask/wqflask/static/new/javascript/search_results.js
@@ -25,9 +25,8 @@ change_buttons = function(check_node = 0) {
 };
 
 $(function() {
-  var add, checked_traits, deselectAll, invert, selectAll;
+  var selectAll, deselectAll, invert;
 
-  checked_traits = null;
   selectAll = function() {
     table_api = $('#trait_table').DataTable();
 
@@ -187,10 +186,6 @@ $(function() {
       href: "/collections/add?hash=" + traits_hash
     });
 
-  };
-
-  removedTraits = function() {
-    return checked_traits.closest("tr").fadeOut();
   };
 
   submitBnw = function() {
--- a/wqflask/wqflask/static/new/javascript/search_results.js
+++ b/wqflask/wqflask/static/new/javascript/search_results.js
@@ -130,6 +130,13 @@ $(function() {
     })
   }
 
+  $(window).keydown(function(event){
+    if((event.keyCode == 13)) {
+      event.preventDefault();
+      return false;
+    }
+  });
+
   $('#select_top').keyup(function(event){
     if (event.keyCode === 13) {
       filter_by_index()
--- a/wqflask/wqflask/static/new/javascript/search_results.js
+++ b/wqflask/wqflask/static/new/javascript/search_results.js
@@ -102,7 +102,13 @@ $(function() {
         try {
           start_index = parseInt(index_set.split("-")[0]);
           end_index = parseInt(index_set.split("-")[1]);
-          for (index = _j = start_index; start_index <= end_index ? _j <= end_index : _j >= end_index; index = start_index <= end_index ? ++_j : --_j) {
+
+          // If start index is higher than end index (for example is the string "10-5" exists) swap values so it'll be interpreted as "5-10" instead
+          if (start_index > end_index) {
+            [start_index, end_index] = [end_index, start_index]
+          }
+
+          for (index = start_index; index <= end_index; index++) {
             index_list.push(index);
           }
         } catch (_error) {
--- a/wqflask/wqflask/static/new/javascript/search_results.js
+++ b/wqflask/wqflask/static/new/javascript/search_results.js
@@ -25,10 +25,10 @@ change_buttons = function(check_node = 0) {
 };
 
 $(function() {
-  var add, checked_traits, deselect_all, invert, remove, removed_traits, select_all;
+  var add, checked_traits, deselectAll, invert, selectAll;
 
   checked_traits = null;
-  select_all = function() {
+  selectAll = function() {
     table_api = $('#trait_table').DataTable();
 
     check_cells = table_api.column(0).nodes().to$();
@@ -44,7 +44,7 @@ $(function() {
     change_buttons();
   };
 
-  deselect_all = function() {
+  deselectAll = function() {
     table_api = $('#trait_table').DataTable();
 
     check_cells = table_api.column(0).nodes().to$();
@@ -93,7 +93,7 @@ $(function() {
       $('#trait_table').DataTable().search($(this).val()).draw();
   });
 
-  parse_index_string = function(idx_string) {
+  parseIndexString = function(idx_string) {
     index_list = [];
     _ref = idx_string.split(",");
     for (_i = 0; _i < _ref.length; _i++) {
@@ -117,9 +117,9 @@ $(function() {
     return new Set(index_list)
   }
 
-  filter_by_index = function() {
+  filterByIndex = function() {
     index_string = $('#select_top').val()
-    index_set = parse_index_string(index_string)
+    index_set = parseIndexString(index_string)
 
     table_api = $('#trait_table').DataTable();
     check_nodes = table_api.column(0).nodes().to$();
@@ -139,15 +139,15 @@ $(function() {
 
   $('#select_top').keyup(function(event){
     if (event.keyCode === 13) {
-      filter_by_index()
+      filterByIndex()
     }
   });
 
   $('#select_top').blur(function() {
-    filter_by_index()
+    filterByIndex()
   });
 
-  add_to_collection = function() {
+  addToCollection = function() {
     var traits;
     table_api = $('#trait_table').DataTable();
     check_nodes = table_api.column(0).nodes().to$();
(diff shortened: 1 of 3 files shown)


wqflask/wqflask/static/new/javascript/search_results.js
```diff
@@ -93,48 +93,69 @@ $(function() {
...
-              check_rows[i].classList.remove("selected")
-            }
-          }
-      }
-      else {
-        for (let i = 0; i < check_rows.length; i++){
-          check_cells[i].childNodes[0].checked = false;
-          if (check_rows[i].classList.contains("selected")){
-            check_rows[i].classList.remove("selected")
+  parse_index_string = function(idx_string) {
```
BonfaceKilz: *nitpick*: Random note on naming things in JavaScript, names should be camelCased instead of snake_cased. so `parse_index_string` would be `parseIndexString`.

wqflask/wqflask/static/new/javascript/search_results.js
```diff
@@ -93,48 +93,69 @@ $(function() {
...
+    index_list = [];
+    _ref = idx_string.split(",");
+    for (_i = 0, _len = _ref.length; _i < _len; _i++) {
```
BonfaceKilz: *issue*: Avoid creating unnecessary variables:

```suggestion
    for (_i = 0; _i < _ref.length; _i++) {
```

wqflask/wqflask/static/new/javascript/search_results.js
```diff
@@ -93,48 +93,69 @@ $(function() {
...
+      index_set = _ref[_i];
+      if (index_set.indexOf('-') !== -1) {
+        try {
+          start_index = parseInt(index_set.split("-")[0]);
+          end_index = parseInt(index_set.split("-")[1]);
+          for (index = _j = start_index; start_index <= end_index ? _j <= end_index : _j >= end_index; index = start_index <= end_index ? ++_j : --_j) {
```
BonfaceKilz: *nitpick:*  This is not readable; and hard to understand what's going.  FWIW, no need to re-write; but let's prefer, even if it's JavaScript, to write things which you can follow.

wqflask/wqflask/static/new/javascript/search_results.js
```diff
@@ -93,48 +93,69 @@ $(function() {
...
+            index_list.push(index);
           }
+        } catch (_error) {
+          error = _error;
+          alert("Syntax error");
```
BonfaceKilz: *question: nitpick:* Is there a way to check for a parsing error other than raising an alert?  Is this expected.  We can also make sure that the values passed to this script are always correct; thereby ridding the need have this `try ... catch` block...

BonfaceKilz (commented):
@zsloan I just added a few minor comments.  Also, let's see if we can improve our commits.  We should pair one of these fine days :)

zsloan:
@BonfaceKilz Thanks for the comments; going to fix/change most of the things you mentioned in a moment. 

Regarding the try/except, that function basically accepts a string meant to indicate which row indices to select. So someone can type something like "1, 5, 10-50, 60" and it will select those rows (so for that string it would select 44 rows). Currently it splits by comma (since that separates the different indices or index ranges) and attempts to parse each section as an index or name of index (so doesn't work if you throw some letters or other characters that aren't either numbers or the "-" character). I'm going to at least change the alert to something more descriptive, but I'm not sure how else to handle that other than directly checking for characters that aren't numbers or the "-" character (which seems like unnecessary overhead, though maybe it's not a big deal).

The reason why that for-loop is so weird (and I'm going to change it to move the logic inside of the for-loop because, as you mentioned, it's really confusing to read) is that it's also trying to account for situations where the user puts, for example,"5-1" (to make it then interpret that as "1-5" instead, by switching the start/end index).

As for the commits, can you give some examples? Are you referring to the commit text?

zsloan:
@BonfaceKilz I ended up replacing the try/catch with a regex pattern match that seems to work correctly (though I wouldn't be surprised if there's something I missed, since I don't have much experience with regex).

BonfaceKilz:

> @BonfaceKilz Thanks for the comments; going to fix/change most of the things you mentioned in a moment. 
>

Sure!

> Regarding the try/except, that function basically accepts a string meant to indicate which row indices to select. So someone can type something like "1, 5, 10-50, 60" and it will select those rows (so for that string it would select 44 rows). Currently it splits by comma (since that separates the different indices or index ranges) and attempts to parse each section as an index or name of index (so doesn't work if you throw some letters or other characters that aren't either numbers or the "-" character). I'm going to at least change the alert to something more descriptive, but I'm not sure how else to handle that other than directly checking for characters that aren't numbers or the "-" character (which seems like unnecessary overhead, though maybe it's not a big deal).
>
Thanks for looking in to this.

> The reason why that for-loop is so weird (and I'm going to change it to move the logic inside of the for-loop because, as you mentioned, it's really confusing to read) is that it's also trying to account for situations where the user puts, for example,"5-1" (to make it then interpret that as "1-5" instead, by switching the start/end index).

I see.  Usually, when writing code, it's best to
have a hack that works; and then later, you
refactor that hack to be more readable.

> As for the commits, can you give some examples? Are you referring to the commit text?
>

Yes.  Look at this example from Arun (in GN3):

```
9a35cdd README: Update link to CI badge.
fa227ea db: Assist mypy with explicit type declaration.
36cc8b1 db: Fix sparql_query return type.
deca94c db: Fix wrong continued indentation.
8b94567 pylintrc: Disable fixme check.
4f8042a Remove pull request template.
85405fe Remove redundant check on the Pearson correlation coefficient.

```

vs:

```
77099ca Fixed when model is set to account for situations with no covariates
baa801e Incorporated Danny's code for calculating QTL main effects into rqtl_wrapper.R
b815236 Fix covariates as described by Danny
00be769 Add covarstruct as a possible keyword for the rqtl_wrapper.R script (not integrated into the script yet though)
79d19f4 Fix issue that caused R/qtl to not use the first covariate when using stratified permutations + covariates
cbb8029 Check if corr_coefficient is NaN, since apparently it's stored as NaN instead of None when it can't be calculcated (which was messing up sorting); it may also be okay to remove the None check, but leaving it for now (#28)
f7becfa Fixed another error where test_generate_rqtl_command didn't include the filename argument (not sure why running unit tests locally doesn't detect this)
84790de Fixed issue where all bool kwargs were always being passed to generate_rqtl_cmd and also made code check if output file already exists (so caching works)
9b8fb0f Added option for enabling permutation strata
0bfe299 Add rqtl_wrapper.R script and necessary imports to guix.scm
bf0a8cb Fix R/qtl command and the way keyword arguments are passed
```

Pointers:

- Make commits in the imperative voice
- Make them descriptive
- Make them brief(less than 80 chars).  A good editor will help you with this.

Why?

- I normally review(or at least I'm getting into
  the groove of reviewing gn2/3 work at the end of
  my day).  Short commits give a good idea of
  what's going on; and makes it a nice experience.

- Git Bisects.  When tracking down bugs, I've
  done my own share of bisects to see where a bug
  was introduced.  Descriptive commits help in
  identifying guesses on where to start bisects.

As an example, this:

```
cbb8029 Check if corr_coefficient is NaN, since apparently it's stored as NaN instead of None when it can't be calculcated (which was messing up sorting); it may also be okay to remove the None check, but leaving it for now (#28)
```

Could be written as:

```
9a35cdd Check if `corr_coefficient` is NaN

The `corr_coefficient` can bi stored as NaN
instead of None when it can't be calculated.  NaN
values messes up sorting.  It's also okay to
remove the None check, but for now I'm leaving it
for now.

#28
```

The subject is brief and descriptive; the body
adds more details should you feel you need to
provide more context.

Also, IMHO this is one of those things that come
with intentional practice, a bit of guidance and
help.  I'm also learning things on the fly ;).  I
hope this brief exposition made sense.

wqflask/wqflask/static/new/javascript/search_results.js
```diff
@@ -25,10 +25,10 @@ change_buttons = function(check_node = 0) {
 };
 
 $(function() {
-  var add, checked_traits, deselect_all, invert, remove, removed_traits, select_all;
+  var add, checked_traits, deselectAll, invert, selectAll;
```
BonfaceKilz: *nitpick:* `checked_traits` -> `checkedTraits`.

*issue:* For new JS code, let's try using `let` or `const` as opposed `var` since all our users will most definitely be using new browsers.  Consider:
```
{
var x = 1;
}
console.log(x)
```
you get: 1
vs 

```
{
var y = 1;
}
console.log(y)
```
you get: `undefined`

`var` is not block-scoped.  

You can read more [here](https://janakachathuranga.medium.com/why-we-use-const-and-avoid-var-3528ee550fb4)

wqflask/wqflask/static/new/javascript/search_results.js
```diff
@@ -93,22 +93,37 @@ $(function() {
...
 
-  parse_index_string = function(idx_string) {
+  /**
+   * parseIndexString takes a string consisting of integers,
+   * hyphens, and/or commas to indicate range(s) of indices
+   * to select a rows and returns the corresponding set of indices
+   * For example - "1, 5-10, 15" would return a set of 8 rows
+   * @return {Set} The list of indices as a Set
+   */
+  parseIndexString = function(idx_string) {
```
BonfaceKilz, line 102: *praise:* Thanks for adding this in.

BonfaceKilz (approved):
This LGTM.

zsloan:
Thanks for the comments on commit messages; that's something very straight-forward to keep in mind (basically making the initial message short and adding more details below when necessary).

I made a couple changes and fixed/added highlighting, but will go ahead and merge now.

PR #661, opened 2022-01-26
Fix/consider within 3dp values on data upload

BonfaceKilz:
#### Description

Fixed precision point errors.  During file upload, excel appended
extra values; so instead of `BXD1,18.8,x,x`, you'd have
`BXD1,18.8000001,20,x`.  There are 2 parts to fixing this:

- During generating the JSON file(when creating the diff), ignore
  modifications where |ε| > 0.001.

- When listing diffs, ignore modification entries where |ε| > 0.001.

For this fix, the second bit was ignored, because it'll lead to
unnecessarily complicated code, given that only one person has really
been testing the system.

N/A

N/A

https://github.com/genenetwork/gn-gemtext-threads/commit/eb435b76b3433ed97837299329229e919c3fffd6

N/A

#### Questions

N/A

### DIFF
--- a/wqflask/wqflask/metadata_edits.py
+++ b/wqflask/wqflask/metadata_edits.py
@@ -257,9 +257,29 @@ def update_phenotype(dataset_id: str, name: str):
         r = run_cmd(cmd=("csvdiff "
                          f"'{uploaded_file_name}' '{new_file_name}' "
                          "--format json"))
+        json_data = json.loads(r.get("output"))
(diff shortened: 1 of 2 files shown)


BonfaceKilz:
@zsloan I'm going to merge this.  You can have a look at it.  For this PR, I've opted to ignore existing diffs, since that'll have us have more hacky code for little value considering we've only had one tester.

PR #663, opened 2022-01-28
Individuals/samples genotypes converter

zsloan:
This adds a script that takes a source .geno file (or JSON file containing a list of .geno files) and a target dummy .geno file (needed just for the ordered strain list) and generated actual .geno files for the target group corresponding with the source .geno file(s).

In the immediate, this is meant to address the need to create a BXD-Longevity .geno file corresponding with all 3 of the "main" BXD group .geno files. But it will also be used to generate genotype files for any other groups of individuals/samples (and maybe later to automatically generate such files when the source .geno file is updated).


wqflask/maintenance/gen_ind_genofiles.py
```diff
@@ -0,0 +1,117 @@
+# Example command: env GN2_PROFILE=/usr/local/guix-profiles/gn-latest-20220122 TMPDIR=/export/local/home/zas1024/gn2-zach/tmp WEBSERVER_MODE=DEBUG LOG_LEVEL=DEBUG SERVER_PORT=5002 GENENETWORK_FILES=/export/local/home/zas1024/gn2-zach/genotype_files SQL_URI=mysql://webqtlout:webqtlout@localhost/db_webqtl ./bin/genenetwork2 ./etc/default_settings.py -c ./maintenance/gen_ind_genofiles.py
```
BonfaceKilz: **nitpick**: Try wrapping lines (if feasible) around 80 chars.  That way, it's easier to read, even in a browser.  Also, if possible, try making the first line: "/usr/bin/env python3" and then use doc-strings for actual documentation.  Here's an example from scripts/authentication/group.py:


```
#!/usr/bin/env python3
"""A script for adding users to a specific group.

Example:

Assuming there are no groups and 'test@bonfacemunyoki.com' does not
exist in Redis:

.. code-block:: bash
  python group.py -n "editors" \
                  -m "me@bonfacemunyoki.com"

results in::

  Successfully created the group: 'editors'
  `HGET groups 0360449f-5a96-4940-8522-b22d62085da9`: {'name': 'editors', 'admins': [], 'members': ['8ad942fe-490d-453e-bd37-56f252e41603'], 'changed_timestamp': 'Oct 19 2021 09:34AM', 'created_timestamp': 'Oct 19 2021 09:34AM'}
  

Assuming we have a group's unique id:

.. code-block:: bash
  python group.py -n "editors" \
                  -m "me@bonfacemunyoki.com" \
                  -g "8ad942fe-490d-453e-bd37-56f252e41603"

now results in::

  Successfully created the group: 'editors'
  `HGET groups 8ad942fe-490d-453e-bd37-56f252e41603`: {'name': 'editors', 'admins': [], 'members': ['8ad942fe-490d-453e-bd37-56f252e41603'], 'changed_timestamp': 'Oct 19 2021 09:38AM', 'created_timestamp': 'Oct 19 2021 09:38AM'}
  
If 'me@bonfacemunyoki.com' exists in 'users' in Redis for the above
command and we run:

.. code-block:: bash
  python group.py -n "editors" \
                  -m "me@bonfacemunyoki.com" \
                  -g "8ad942fe-490d-453e-bd37-56f252e41603"

now results in::

  No new group was created.
  `HGET groups 8ad942fe-490d-453e-bd37-56f252e41603`: {'name': 'editors', 'admins': [], 'members': ['8ad942fe-490d-453e-bd37-56f252e41603'], 'changed_timestamp': 'Oct 19 2021 09:40AM'}


"""
...
```

wqflask/maintenance/gen_ind_genofiles.py
```diff
@@ -0,0 +1,117 @@
...
+                    'Mb': line_items[header_columns.index("Mb")],
+                    'cM': line_items[header_columns.index("cM")],
+                    'genotypes': list(zip(sample_list, [item.strip() for item in line_items][geno_start_col:]))
+                }
+                marker_genotypes.append(this_marker)
+
+    return marker_genotypes
+            
+if __name__ == "__main__":
+    print("command line arguments:\n\t%s" % sys.argv)
```
BonfaceKilz: **nitpick:** Try using f-strings for string substitutions.

wqflask/maintenance/gen_ind_genofiles.py
```diff
@@ -0,0 +1,117 @@
...
+    # Generate the output .geno files
+    generate_new_genofiles(strain_genotypes(source_genofile), target_groups)
+
+def group_samples(target_group: str) -> List:
+    """
+    Get the group samples from its "dummy" .geno file (which still contains the sample list)
+    """
+
+    # Allow for inputting the target group as either the group name or .geno file
+    file_location = app.config.get("GENENETWORK_FILES") + "/genotype/" + target_group
```
BonfaceKilz: **issue: question** This being a script, why are we using "app"?  Shouldn't we pass in things through the cli?

wqflask/maintenance/gen_ind_genofiles.py
```diff
@@ -0,0 +1,117 @@
...
+            if line[0] in ["#", "@"] or not len(line):
+                continue
+    
+            line_items = line.split("\t")
+            sample_list = [item for item in line_items if item not in ["Chr", "Locus", "Mb", "cM"]]
+            break
+
+    return sample_list
+
+def strain_genotypes(strain_genofile: str) -> List:
```
BonfaceKilz, line 181: **nitpick:** Given the doc-strings, this function name could be better.   Generally, functions in Python are verbs; and classes are nouns.  So this would be something like: `read_genotypes_from_file`

wqflask/maintenance/gen_ind_genofiles.py
```diff
@@ -0,0 +1,117 @@
...
+            'Locus': 'marker1',
+            'Mb': '10.0',
+            'cM': '8.0',
+            'genotypes': [('BXD1', 'B'), ('BXD2', 'D'), ('BXD3', 'H'), ...]
+        },
+        ...
+    ]
+    """
+
+    file_location = app.config.get("GENENETWORK_FILES") + "/genotype/" + strain_genofile
```
BonfaceKilz: **issue:** See above comment on editing "app".

BonfaceKilz (commented):
Generally, if you can, you should limit functions to doing only one and only one thing.

wqflask/maintenance/gen_ind_genofiles.py
zsloan: Ah yeah, this is easy for me to forget because I use an ultra-wide monitor. I'll see if I can add something to my IDE that will remind me (I think there's some setting for this).

I'll also just put that information in a docstring.

wqflask/maintenance/gen_ind_genofiles.py
zsloan: This line is actually going to be removed and probably shouldn't have been included in the commit (which I only made once I was sure that what I had was at least partially working.

wqflask/maintenance/gen_ind_genofiles.py
zsloan: This script is being run in the way described here - https://github.com/genenetwork/genenetwork2/blob/6fa39b4596354fdf34201b4d7e10e05e6167ec6b/bin/genenetwork2#L24-L28

But I agree that it's probably more simple to make it "stand-alone" and just require that certain environment variables be set (or passed as arguments I guess), so I'll change it do one (or optionally both I guess) of those things.

wqflask/maintenance/gen_ind_genofiles.py
zsloan: I actually was thinking of that one comment @arunisaac had made about function naming when naming this function (in terms of the function just returning something instead of performing an operation). I normally would have gone with something like you mention, but was thinking along the lines of "it's just returning this specific thing instead of changing something's state."

BonfaceKilz:

> @zsloan commented on this pull request.
>
>
>
>> +    # containing a set of .geno files (and their corresponding file names and sample lists)
> +    if ".json" in args[2]:
> +        target_groups = json.load(args[2])['genofile']
> +    else:
> +        target_groups = [args[2]]
> +
> +    # Generate the output .geno files
> +    generate_new_genofiles(strain_genotypes(source_genofile), target_groups)
> +
> +def group_samples(target_group: str) -> List:
> +    """
> +    Get the group samples from its "dummy" .geno file (which still contains the sample list)
> +    """
> +
> +    # Allow for inputting the target group as either the group name or .geno file
> +    file_location = app.config.get("GENENETWORK_FILES") + "/genotype/" + target_group
>
> This script is being run in the way described here - https://github.com/genenetwork/genenetwork2/blob/6fa39b4596354fdf34201b4d7e10e05e6167ec6b/bin/genenetwork2#L24-L28
>
> But I agree that it's probably more simple to make it "stand-alone" and just require that certain environment variables be set (or passed as arguments I guess), so I'll change it do one (or optionally both I guess) of those things.
>
I agree

BonfaceKilz:

> @zsloan commented on this pull request.
>
>
>
>> +        file_location += ".geno"
> +
> +    sample_list = []
> +    with open(file_location, "r") as target_geno:
> +        for i, line in enumerate(target_geno):
> +            # Skip header lines
> +            if line[0] in ["#", "@"] or not len(line):
> +                continue
> +    
> +            line_items = line.split("\t")
> +            sample_list = [item for item in line_items if item not in ["Chr", "Locus", "Mb", "cM"]]
> +            break
> +
> +    return sample_list
> +
> +def strain_genotypes(strain_genofile: str) -> List:
>
> I actually was thinking of that one comment
> @arunisaac had made about function naming when
> naming this function (in terms of the function
> just returning something instead of performing
> an operation).

I think I remember this.  And if I recall
correctly, it was a very lisp-y remark.  See this
discussion:
<https://softwareengineering.stackexchange.com/questions/334135/how-should-i-name-functions-that-return-values-in-pythonhttps://softwareengineering.stackexchange.com/questions/334135/how-should-i-name-functions-that-return-values-in-python>

>I normally would have gone with something like you mention, but was thinking along the lines of "it's just returning this specific thing instead of changing something's state."
>
I see.

BonfaceKilz (approved):
Nothing seems off to me for. Thanks for this!

pjotrp:
Nice code!

PR #674, opened 2022-02-28
Fix fulltext search for certain searches

zsloan:
Hyphens in fulltext searches were causing problems, but a recent commit
I made to fix the issue (6496f75af521d6801c25de682d2852e14d597fed) apparently caused some issues for other types
of searches, in addition to make such searches somewhat slower.

Apparently the issue wasn't just the hyphens, but also the text to
either side of the hyphen being lower than the minimum word length
(which is either 2 or 3 for us, can't remember). To try and address
this, I did a regular expression check for a pattern with text of <3
legnth to either side of a hyphen, and when that's the case I add quotes
from the search term plus an asterisk, which seems to be necessary to
get it to not treat the hyphen as a delimiter and to correctly detect
the search term during a boolean fulltext search.

### DIFF
--- a/wqflask/wqflask/do_search.py
+++ b/wqflask/wqflask/do_search.py
@@ -1,6 +1,7 @@
-import string
-import requests
 import json
+import re
+import requests
+import string
 
 from flask import Flask, g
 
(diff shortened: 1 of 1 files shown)


arunisaac:
> Hyphens in fulltext searches were causing problems, but a recent commit
> I made to fix the issue (6496f75af521d6801c25de682d2852e14d597fed)

Could you also add a test case, if you haven't already, to ensure that
we have no regressions on this issue?

Thanks!

zsloan:
Yeah, I'm going to add one similar to the one here (that can check number
of results) tomorrow -
https://github.com/genenetwork/genenetwork2/blob/testing/test/requests/main_web_functionality.py

I'm not sure how to make one more specific than that (I'm sure it's
possible, but I don't know how without doing something weird like comparing
all the HTML), but that should at least be enough to verify that nothing
has changed for one or two different searches relevant to this issue.

BonfaceKilz:

> Yeah, I'm going to add one similar to the one here (that can check number
> of results) tomorrow -
> https://github.com/genenetwork/genenetwork2/blob/testing/test/requests/main_web_functionality.py
>
> I'm not sure how to make one more specific than that (I'm sure it's
> possible, but I don't know how without doing something weird like comparing
> all the HTML), but that should at least be enough to verify that nothing
> has changed for one or two different searches relevant to this issue.

That's an E2E test though; and we'd have to
resurrect Mechanical Rob.

PR #676, opened 2022-03-01
Add some search tests to check some fulltext searches

zsloan:
This PR adds some search tests, mainly to ensure that the fulltext searches continue to work the same way. It only checks for the correct number of results, but that should be sufficient if done across multiple possible searches.

Not sure how to actually make this test work with the CI, or whether it's even in the right place.

### DIFF
--- a/test/requests/search_tests.py
+++ b/test/requests/search_tests.py
@@ -1,0 +1,45 @@
+import requests
+
+searches = [
+    { 
+        "found_text": "1 records were found",
+        "data": {
+            "species": "mouse",
+            "group": "BXD",
+            "type": "Hippocampus mRNA",
+            "dataset": "HC_M2_0606_P",
+            "search_terms_or": "PD-1",
+            "search_terms_and": ""
+        }
+    },
+    { 
+        "found_text": "4 records were found",
+        "data": {
+            "species": "mouse",
+            "group": "BXD",
+            "type": "Hippocampus mRNA",
+            "dataset": "HC_M2_0606_P",
+            "search_terms_or": "Ank-1",
+            "search_terms_and": ""
+        }
+    },
+    { 
+        "found_text": "1017 records were found",
+        "data": {
+            "species": "mouse",
+            "group": "BXD",
+            "type": "Hippocampus mRNA",
+            "dataset": "HC_M2_0606_P",
+            "search_terms_or": "sh*",
+            "search_terms_and": ""
+        }
+    }
+]
+
+def check_fulltext_searches(host):
+    for this_search in searches:
+        result = requests.get(host+"/search", params=this_search["data"])
+        found = result.text.find(this_search["found_text"])
+        assert(found >= 0)
+        assert(result.status_code == 200)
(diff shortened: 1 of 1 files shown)


test/requests/search_tests.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+]
+
+def check_fulltext_searches(host):
+    for this_search in searches:
+        result = requests.get(host+"/search", params=this_search["data"])
+        found = result.text.find(this_search["found_text"])
+        assert(found >= 0)
+        assert(result.status_code == 200)
+        print("OK")
```
BonfaceKilz, line 45: **nitpick**: No need for this statement.

test/requests/search_tests.py
```diff
@@ -0,0 +1,45 @@
...
+            "type": "Hippocampus mRNA",
+            "dataset": "HC_M2_0606_P",
+            "search_terms_or": "sh*",
+            "search_terms_and": ""
+        }
+    }
+]
+
+def check_fulltext_searches(host):
+    for this_search in searches:
```
BonfaceKilz, line 40: **nitpick**: Try (same case for me) to be more mindful of naming. `this_search` can be off. How about:
```suggestion
    for search in searches:

```

BonfaceKilz (commented):
Also for testing, try to have these tests pass on your own local set-up. Then, add then try to capture the tests that were failing. Happy Hacking!

zsloan:
How exactly do I go about doing a test like this? I more or less understand
how to run unit tests (and have existing examples of those to look at), but
I'm not sure how to run this kind of test (the only examples I could find
are the ones in the directory I mentioned in the other email, which as you
mentioned I think depended on Mechanical Rob).

BonfaceKilz:

> How exactly do I go about doing a test like this? I more or less understand
> how to run unit tests (and have existing examples of those to look at), but
> I'm not sure how to run this kind of test (the only examples I could find
> are the ones in the directory I mentioned in the other email, which as you
> mentioned I think depended on Mechanical Rob).
>
Can you run Mechanical Rob? If not, then we need
to make that clear for both you and the rest of
the team. That's on my plate. Also, in addition to
running Mechanical Rob, you could inspect the
underlying functions in that global search and
test for those. Some tests could be to check that
the right SQL query is being run; or that the
processing of results from said query is done
properly. That said, in gn2 some functions are
notoriously difficult to test, so don't spend too
much time on that. Also, I've found that using
pudb for class, function and variable
introspection has been really useful in both
troubleshooting and working out how to test. When
we meet together, I could briefly show you how to
use that.

arunisaac:
> Not sure how to actually make this test work with the CI, or whether
> it&#39;s even in the right place.

Currently, the CI will run all tests run by the following command. You
don't have to do anything special to get it into the CI.

```
./bin/genenetwork2 /etc/default_settings.py -c -m unittest -v
```

End-to-end tests with Mechanical Rob are not yet implemented in the
CI. I am in conversation with Bonface about that.

PR #679, opened 2022-03-03
Fix/caching

Alexanderlacuna:
Pr fixes issue https://github.com/genenetwork/genenetwork2/issues/677.where empty text files results to json decoding error.This
should never happen unless the cache isn't working at all

(TODO)regression tests

### DIFF
--- a/wqflask/wqflask/correlation/pre_computes.py
+++ b/wqflask/wqflask/correlation/pre_computes.py
@@ -6,6 +6,8 @@ from pathlib import Path
 from base.data_set import query_table_timestamp
 from base.webqtlConfig import TMPDIR
 
+from json.decoder import JSONDecodeError
+
 
 def fetch_all_cached_metadata(dataset_name):
     """in a gvein dataset fetch all the traits metadata"""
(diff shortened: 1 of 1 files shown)


zsloan:
Ah, this is a good solution. In this case I don't think the JSON file was empty (it said it had an error at position 600000-something), but this should still resolve the issue of a flawed cache file by removing it and regenerating a new one.

wqflask/wqflask/correlation/pre_computes.py
```diff
@@ -6,29 +6,40 @@
...
 
+from json.decoder import JSONDecodeError
+
 
 def fetch_all_cached_metadata(dataset_name):
     """in a gvein dataset fetch all the traits metadata"""
     file_name = generate_filename(dataset_name, suffix="metadata")
 
-    file_path = os.path.join(TMPDIR, file_name)
+    file_path = Path(TMPDIR, file_name)
```
BonfaceKilz, line 16: **Question**: What's the preference of this over "os.path.join"? Here I'm just being curious.

BonfaceKilz: Actually, I later see that you use the "unlink" and "touch" so :+1:

BonfaceKilz (approved):
Nothing looks off to me. Thanks for your contributions and acting promptly on https://github.com/genenetwork/genenetwork2/issues/677

Alexanderlacuna:
going ahead to merge this

PR #701, opened 2022-04-28
Feature/collection export

zsloan:
This PR adds the option to export/import collections (or any trait list, such as search results).

The output file just contains identifying information for each trait (record ID + short dataset name). This way users can share collections with one another.

The current implementation might not work with "temp" traits.


zsloan:
This appears to have already been merged, separately from this PR for some reason, so closing it.

PR #706, opened 2022-05-19
Feature/edit case attrs in new page

BonfaceKilz:
#### Description

This PR adds CRUD operations on case-attributes with an approval step. Approve/rejected diffs related to case-attributes are stored in "caseattributes_audit.sql" table. This table should be created before this feature gets in use.

https://issues.genenetwork.org/issues/case-attr-edit-error.html


BonfaceKilz:
@zsloan when you get the bandwidth, could you create this table: https://github.com/genenetwork/genenetwork3/blob/main/sql/update/caseattributes_audit.sql ?

PR #715, opened 2022-07-04
Chore/GN2 Clean up

BonfaceKilz:
#### Description
This PR:
- Removes FlaskDebugToolbar.  It's utility is already provided natively in Python's pudb module. @fredmanglis @Alexanderlacuna and @zsloan take note.  Eventually, I'll remove this as an input in gn2's definition in our channel.
- Uses new function signature when fetching case-attributes.
- Fix edge-case when listing diffs if a user who edits a file does not exist in Redis.  This should never happen, but nevertheless, I catered for it.

### DIFF
--- a/wqflask/runserver.py
+++ b/wqflask/runserver.py
@@ -23,9 +23,7 @@ app_config()
 werkzeug_logger = logging.getLogger('werkzeug')
 
 if WEBSERVER_MODE == 'DEBUG':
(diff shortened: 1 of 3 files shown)


zsloan:
Good idea with addressing the edge case of a user not existing in Redis. Issues similar to this popped up in the past (due to a mistake in the code, but still - users who created their accounts at different times sometimes didn't have certain attributes, or had multiple accounts for the same e-mail address).

PR #720, opened 2022-08-18
Chores/gn2 queries optimization

Alexanderlacuna:
optimization of queries that fetch correlation data,traits metadata

### DIFF
--- a/wqflask/wqflask/correlation/rust_correlation.py
+++ b/wqflask/wqflask/correlation/rust_correlation.py
@@ -15,6 +15,34 @@ from gn3.computations.rust_correlation import parse_tissue_corr_data
 from gn3.db_utils import database_connector
 
 
+def query_probes_metadata(dataset, results):
+
+    """query traits metadata in bulk for probeset"""
+
+    with database_connector() as conn:
+        with conn.cursor() as cursor:
+
+            query = """
+                    SELECT ProbeSet.Name,ProbeSet.Chr,ProbeSet.Mb,
+                    Symbol,mean,description,additive,LRS,Geno.Chr, Geno.Mb
+                    from Geno, Species,ProbeSet,ProbeSetXRef,ProbeSetFreeze
+                    where ProbeSet.Name in ({}) and
+                    Species.Name = %s and
+                    Geno.Name = ProbeSetXRef.Locus and
+                    Geno.SpeciesId = Species.Id and
+                    ProbeSet.Id=ProbeSetXRef.ProbeSetId and
+                    ProbeSetFreeze.Id = ProbeSetXRef.ProbeSetFreezeId and
+                    ProbeSetFreeze.Name = %s
+                """.format(", ".join(["%s"] * len(trait_list)))
+
+            cursor.execute(query,
+                           (tuple(trait_list) +
+                            (dataset.group.species,) + (dataset.name,))
+                           )
+
+            return cursor.fetchall()
+
+
 def chunk_dataset(dataset, steps, name):
 
     results = []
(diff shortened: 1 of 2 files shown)


fredmanglis (commented):
In [line 36](https://github.com/genenetwork/genenetwork2/pull/720/commits/7d089d58d9e62fba6d61f17f93148054cf7e6730#diff-746045489e3953c7c77c26c24e39c9afa677da97dd74c86296626d4f24e05043R36) and [line 39](https://github.com/genenetwork/genenetwork2/pull/720/commits/7d089d58d9e62fba6d61f17f93148054cf7e6730#diff-746045489e3953c7c77c26c24e39c9afa677da97dd74c86296626d4f24e05043R39), you make reference to the `trait_list` variable, but there is no declaration of the variable anywhere: even in the [function arguments](https://github.com/genenetwork/genenetwork2/pull/720/commits/7d089d58d9e62fba6d61f17f93148054cf7e6730#diff-746045489e3953c7c77c26c24e39c9afa677da97dd74c86296626d4f24e05043R18).

*nitpick*: Also, maybe remove all those empty lines - they serve no purpose, and make reading the code more complex.

fredmanglis (commented):
Rather than a for loop, why don't you use a dict comprehension for [this region](https://github.com/genenetwork/genenetwork2/pull/720/commits/c3d95c1f81d35b17ea8aa027d7d2973d5bc1d676#diff-746045489e3953c7c77c26c24e39c9afa677da97dd74c86296626d4f24e05043R53-R71)

Something like:

```python
def get_metadata(dataset, traits):
    "Function documentation goes here ..."
    return {
        trait_name: {
            "name": trait_name,
            "view": True,
            "symbol": symbol,
            ...
        }
        for trait_name, probe_chr, probe_mb, ...
        in query_probes_metadata(dataset, traits)
    }
```

That way, you avoid having unnecessary names (the `db_results` and `metadata` names) and you also get rid of all the extraneous spaces.

You can do this in a new commit, then use **git rebase** to replace commit c3d95c1f81d35b17ea8aa027d7d2973d5bc1d676 with the new commit...

If unsure how, I am willing to guide you through that, and while I cannot speak for @BonfaceKilz I get the impression they will probably not mind walking you through that either.

fredmanglis (commented):
**teachable moment**: If you had committed [these cleanup changes](https://github.com/genenetwork/genenetwork2/pull/720/commits/1aa6e399a07eb5dfe1abe224cf0d55d9e0c3216d#diff-746045489e3953c7c77c26c24e39c9afa677da97dd74c86296626d4f24e05043R56-R69) on their own separate commit, you could then have used the **fixup** feature of **git rebase** to "merge" the new cleanup commit with the older one, keeping the older commit's message.

It makes for cleaner commits.

fredmanglis:
Regarding the "cleanup changes" I mentioned above, I got curious and read up some more on git rebase - It turns out you can [split a commit into multiple commits](https://www.internalpointers.com/post/split-commit-into-smaller-ones-git).

Alexanderlacuna:
@fredmanglis  This can be merged as it is.

fredmanglis:
@Alexanderlacuna Merged!!

BonfaceKilz:
Muriithi Frederick Muriuki
***@***.***> aliandika:

[...]

> You can do this in a new commit, then use **git rebase** to replace commit c3d95c1f81d35b17ea8aa027d7d2973d5bc1d676 with the new commit...
>
> If unsure how, I am willing to guide you through that, and while I cannot speak for @BonfaceKilz I get the impression they will probably not mind walking you through that either.
>

I'm down with this.  Though I think ATM Alex liase
with Fred since you are working on similar things
;)


-- 
(To agree without understanding is inane.
   To disagree without understanding is impudent.)
    (hkp://keys.openpgp.org))

PR #731, opened 2022-09-08
Feature/select samples export

zsloan:
This just adds the option for a user to filter the trait page export by selected rows. If no rows are selected, it just exports everything.

It also renames some variables to camelCase and removes some logging.

### DIFF
--- a/wqflask/wqflask/static/new/javascript/show_trait.js
+++ b/wqflask/wqflask/static/new/javascript/show_trait.js
@@ -638,11 +638,11 @@ populateSampleAttributesValuesDropdown = function() {
   $('#attribute_values').empty();
   sample_attributes = [];
 
-  var attributes_as_list = Object.keys(js_data.attributes).map(function(key) {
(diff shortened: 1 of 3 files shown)


wqflask/wqflask/static/new/javascript/show_trait.js
```diff
@@ -480,7 +480,7 @@ fetchSampleValues = function() {
 }
 
 editDataChange = function() {
-  var already_seen, checkbox, name, real_dict, real_value, real_variance, row, rows, sample_sets, table, tables, _i, _j, _len, _len1;
+  var already_seen, sample_sets, table, tables, _i, _j, _len;
```
BonfaceKilz, line 483: *nitpick:* Prefer to "let" to "var" in order to avoid scoping issues that can be hard to debug.

BonfaceKilz (commented):
@zsloan this looks good to me.  The only other remark I have is: when comparing strings (or anything for that matter), prefer to use the stricter "===" over "==" :smile:

PR #735, opened 2022-09-20
Decompose file into separate modules

fredmanglis:
To ease future refactors on the code, decompose the file into a module with multiple modules that can be refactored semi-independently.

#### Description

This commit splits out the various classes and functions in `wqflask/base/dataset.py` into a package `wqflask/base/dataset` with multiple modules.

https://issues.genenetwork.org/issues/simplify_data_set_py_in_gn2

https://issues.genenetwork.org/issues/simplify_data_set_py_in_gn2


zsloan:
Should this only require changing the way dataset is imported? It seems so from what I can tell.

fredmanglis:
@zsloan actually, not really. Everything should work as is.

fredmanglis:
@zsloan For most part, you should be able to import everything with:

```python
from base.data_set import ...
```

if you look at **/wqflask/base/__init__.py** I imported all the classes and functions from the modules in the new **base.data_set** package

PR #736, opened 2022-09-20
write datasets to text files

Alexanderlacuna:

### DIFF
--- a/wqflask/wqflask/correlation/pre_computes.py
+++ b/wqflask/wqflask/correlation/pre_computes.py
@@ -171,7 +171,7 @@ def get_datasets_data(base_dataset, target_dataset_data):
     return (target_results, base_results)
 
 
-def fetch_text_file(dataset_name, conn, text_dir=TEXTDIR):
+def fetch_text_file(dataset_name, conn, text_dir=TMPDIR):
     """fetch textfiles with strain vals if exists"""
 
     with conn.cursor() as cursor:
@@ -207,7 +207,7 @@ def read_text_file(sample_dict, file_path):
         return (sample_vals, [",".join([line[i] for i in _posit]) for line in csv_reader])
 
 
-def write_db_to_textfile(db_name, conn, text_dir=TEXTDIR):
+def write_db_to_textfile(db_name, conn, text_dir=TMPDIR):
 
     def __generate_file_name__(db_name):
         # todo add expiry time and checker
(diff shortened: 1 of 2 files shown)


Alexanderlacuna:
@fredmanglis this is ready for review the  only  issue is that am storing the text files in /tmp directory which might 
not be that persistent  .GN1 stores the textfiles  in  /gnshare/gn directory am not sure if this is a thing in gn2 maybe  @zsloan

zsloan:
We can just create a directory for these files (in /tmp or elsewhere) and maybe have it passed in as one of the environment variables (like we do the genotype directory)?

fredmanglis (commented):
:+1:

fredmanglis:
> ... the only issue is that am storing the text files in /tmp directory which might not be that persistent  ...

Well, that is not too much of a problem, since the files can be auto-regenerated with your work, right?

A bigger problem is the text-files going stale: if the dataset, or specific traits within a dataset are updated, the data in the file(s) is effectively out of date (I think).

This is a much bigger problem, and I think some of the versioning work that @BonfaceKilz or @arunisaac are working on will come in handy, or might actually be directly related to this.

In any case, I think this is good to go, and the extra concerns can be dealt with by and by.

zsloan:
Yeah, updates are an issue, but fortunately the vast majority of ProbeSet
datasets are never updated (it's more of a problem for phenotypes).

fredmanglis:
@Alexanderlacuna I have merged your work and pushed it.

Alexanderlacuna:
> We can just create a directory for these files (in /tmp or elsewhere) and maybe have it passed in as one of the environment variables (like we do the genotype directory)?

that would  be  better to make  everything well organized

Alexanderlacuna:
> > ... the only issue is that am storing the text files in /tmp directory which might not be that persistent  ...
> 
> Well, that is not too much of a problem, since the files can be auto-regenerated with your work, right?
> 
> A bigger problem is the text-files going stale: if the dataset, or specific traits within a dataset are updated, the data in the file(s) is effectively out of date (I think).
> 
> This is a much bigger problem, and I think some of the versioning work that @BonfaceKilz or @arunisaac are working on will come in handy, or might actually be directly related to this.
> 
> In any case, I think this is good to go, and the extra concerns can be dealt with by and by.

Yeah  if am not wrong the /tmp  directory  is deleted weekly. we  could also solve this by having date of  last modified for tables  as  a hash    included in the filename

PR #739, opened 2022-10-22
fix issue;parsing sample data

Alexanderlacuna:
A fix for issue raised here when parsing all samples:
https://github.com/genenetwork/gn-gemtext-threads/blob/main/issues/correlations-all-samples.gmi


Alexanderlacuna:
merging this,minor fixes.

PR #742, opened 2022-11-07
init code refactoring

Alexanderlacuna:


wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -74,6 +74,193 @@ def set_template_vars(start_vars, correlation_data):
     return correlation_data
 
 
+def apply_filters(trait, target_trait, target_dataset, **filters):
+    def __p_val_filter__(p_lower, p_upper):
+        return not (float(trait.get('corr_coefficient', 0.0)) >= p_lower and
+                    float(trait.get('corr_coefficient', 0.0)) <= p_upper)
```
BonfaceKilz: A more pythonic (and readable way) to write this is:

```suggestion
        return not (float(p_upper >= trait.get('corr_coefficient', 0.0)) >= p_lower)
```

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -74,6 +74,193 @@ def set_template_vars(start_vars, correlation_data):
...
+
+    def __min_filter__(min_expr):
+        if (target_dataset['type'] in ["ProbeSet", "Publish"] and target_trait['mean']):
+            return (min_expr != None) and (float(target_trait['mean']) < min_expr)
```
BonfaceKilz, line 76: Similar comment to above:
```suggestion
            return float(target_trait['mean']) > min_expr != None
```

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -74,6 +74,193 @@ def set_template_vars(start_vars, correlation_data):
...
+        if target_dataset["type"] in ["ProbeSet", "'Geno"] and location_type == "gene":
+
+            return ((location_chr != None and (target_trait["chr"] != location_chr)
+                     or
+                     (min_location_mb != None) and (
+                     float(target_trait['mb']) < min_location_mb)
+                     or
+                     max_location_mb != None) and
+                    (float(target_trait['mb']) > float(max_location_mb)
+                     ))
```
BonfaceKilz: In this case too, you could chain operations together as demo'd above.

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -74,6 +74,193 @@ def set_template_vars(start_vars, correlation_data):
...
+        elif target_dataset["type"] in ["ProbeSet", "Publish"]:
+            return ((location_chr != None and (target_trait["lrs_chr"] != location_chr)
+                     or
+                     (min_location_mb != None) and (
+                         float(target_trait['lrs_mb']) < float(min_location_mb))
+                     or
+                     (max_location_mb != None) and (
+                float(target_trait['lrs_mb']) > float(max_location_mb))
+            )
```
BonfaceKilz: Same sentiment as above.

wqflask/wqflask/correlation/show_corr_results.py
```diff
@@ -86,139 +273,18 @@ def correlation_json_for_table(start_vars, correlation_data, this_trait, this_da
...
-            if bool(target_trait['additive']):
-                results_dict['additive'] = f"{float(target_trait['additive']):.3f}"
-        else:
-            results_dict['location'] = target_trait['location']
 
-        results_list.append(results_dict)
+    dataset_metadata = generate_table_metadata({name for trait in corr_results
+                                                for (name, _val) in trait.items()},
+                                               correlation_data["traits_metadata"],
+                                               target_dataset_ob)
```
BonfaceKilz: This nested "for" loop would be difficult to maintain.  I'd advise using a more simpler, albeit verbose, loop instead.

wqflask/wqflask/correlation/show_corr_results.py
Alexanderlacuna: Awesome

wqflask/wqflask/correlation/show_corr_results.py
Alexanderlacuna: @BonfaceKilz  the above example may lead to unintended results
for example

min_expr = None
``` python
return 2.4 < min_expr is not None

results to 
TypeError: '<' not supported between instances of 'float' and 'str'

```
and if you use bracket that means you will be comparing int to boolean

BonfaceKilz:
Alexander Kabui ***@***.***>
aliandika:

> @Alexanderlacuna commented on this pull request.
>
>
>
>> @@ -74,6 +74,193 @@ def set_template_vars(start_vars, correlation_data):
>      return correlation_data
>  
>  
> +def apply_filters(trait, target_trait, target_dataset, **filters):
> +    def __p_val_filter__(p_lower, p_upper):
> +        return not (float(trait.get('corr_coefficient', 0.0)) >= p_lower and
> +                    float(trait.get('corr_coefficient', 0.0)) <= p_upper)
> +
> +    def __min_filter__(min_expr):
> +        if (target_dataset['type'] in ["ProbeSet", "Publish"] and target_trait['mean']):
> +            return (min_expr != None) and (float(target_trait['mean']) < min_expr)
>
> @BonfaceKilz  the above example may lead to unintended results
> for example
>
> min_expr = None
> ``` python
> return 2.4 < min_expr is not None
>
> results to 
> TypeError: '<' not supported between instances of 'float' and 'str'
>
> ```
> and if you use bracket that means you will be comparing int to boolean

You do have a point.  Thanks for being keen on
this.  That said, we have:

    return 2.4 < min_expr is not None

which is [strictly] not [semantically] the same
as:

    return 2.4 < min_expr != None

It's Worth pointing out that the "is" keyword
checks whether the objects are the same.  You
generally don't want to use this most of the time.
OTOH, "!=" checks for equality of value - which
you want to use most of the time.  Now on to the
important stuff:

Order matters.

Assuming "min_expr" is None, then in ...

    return 2.4 < min_expr != None

... indeed an error, as you correctly pointed out,
will be generated (+1 for spotting that).

However, you can circumvent that using
short-circuiting by checking for the None value
first:

    None != min_expr > 2.4

That said, in this particular case, for
readability's sake, in lieu of your observations,
opt for the simpler [original] case:

   (min_expr != None) and (float(target_trait['mean']) < min_expr)

In conclusion, whichever expression you decide to
use is a matter of taste.  Choose what you see
fit.

-- 
    (hkp://keys.openpgp.org))

Alexanderlacuna:
@zsloan  could try this out and see if there are any issues?
Also check  this out [https://github.com/genenetwork/genenetwork2/pull/746/files](https://github.com/genenetwork/genenetwork2/pull/746/files)

PR #752, opened 2022-11-28
bugfix

Alexanderlacuna:
* this stage is already handled by the rust code  hence doing positional comparison lead to different results than expected
* this is necessary to make sure that the number of items in each row to the number of the x_vals
* a positional comparison is done on the rust code is done to remove nan values
```
//for example
x_vals  = [1.21,1.131,121.11[
y_vals = [Nan,1.1,1.2]
//
parsed_x_val = [1.131,121.11]
parsed_y_val= [1.1,1.2]

```

### DIFF
--- a/wqflask/wqflask/correlation/rust_correlation.py
+++ b/wqflask/wqflask/correlation/rust_correlation.py
@@ -268,7 +268,7 @@ def __compute_sample_corr__(
     target_dataset.get_trait_data(list(sample_data.keys()))
 
     def __merge_key_and_values__(rows, current):
-        wo_nones = [value for value in current[1] if value is not None]
+        wo_nones = [value for value in current[1]]
         if len(wo_nones) > 0:
             return rows + [[current[0]] + wo_nones]
         return rows


zsloan:
I think this might be the source of problem, since when I make this same change on my branch the results seem correct (there's a separate error, but it's one I think I can fix today - there are some None's in the output that are causing an error when converting to JSON).

Alexanderlacuna:
> I think this might be the source of problem, since when I make this same change on my branch the results seem correct (there's a separate error, but it's one I think I can fix today - there are some None's in the output that are causing an error when converting to JSON).

Awesome
In addition to that I have fixed the issue for sorting nan values for rust once merged upstream they will occur less oftenly in the results.
The current issue yet to be fixed is that for index error when reading probeset data from files

zsloan:
After more testing, this appears to fix the results. I'm going to go ahead and merge it since it's a minor change (albeit with major effects), because I want to get the fix on production as soon as possible.

BonfaceKilz:
Alexander Kabui ***@***.***>
anaandika:

Ola ola o/ o/.

Minor comment and very nitpicky.

> -- Commit Summary --
>
>   * this commit remove the stage for parsing rows to remove nan values
>

When writing commits, here's a trick I use:

If this commit is applied this commit will: <your
message in the present tense>

Good commits make it easier to git-bisect, and
also have a rough overview of things.  Please have
a look at everyone else's commits as a reference.

So for the above, I would have written it as:

--8<---------------cut here---------------start------------->8---
Remove None checks when parsing rows

* wqflask/wqflask/correlations/rust_correlation.py
(merge_results)<__merge_key_and_values__>: Delete None check in loop.
--8<---------------cut here---------------end--------------->8---

For more read:
    <https://www.gnu.org/prep/standards/html_node/Change-Logs.html#Change-Logs>

-- 
    (hkp://keys.openpgp.org))

PR #755, opened 2022-11-30
Remove check for shortnames/fullnames in probeset metadata query

BonfaceKilz:
The InfoFiles tables has 2 unique constraints:

```
SELECT DISTINCT CONSTRAINT_NAME FROM information_schema.TABLE_CONSTRAINTS WHERE table_name = 'InfoFiles' and constraint_type = 'UNIQUE';
```
which results in:

```
InfoPageName
GN_AccesionId
```
However, shortname/fullname can have NULL values thereby producing errors in some pages that display the probeset data.


BonfaceKilz:
@zsloan could you merge this as soon as you can?

zsloan:
Thanks for this change; I was actually just about to look at this myself,
since someone had reported an issue with it.

PR #757, opened 2022-12-01
Fix/fix value error when using query sql

BonfaceKilz:
This PR fixes a subtle edge-case where query_sql returns a None results.  This bug prevents some trait pages from showing up.


BonfaceKilz:
@zsloan could you review this and test it out in your branch.  I'll merge this (or you could merge this) once you can confirm it fixes the buy you pointed out :metal:

zsloan:
This works for me and seems to fix the problem! I'll go ahead and merge.

BonfaceKilz:
zsloan ***@***.***> aliandika:

> This works for me and seems to fix the problem! I'll go ahead and merge.

Thanks!

-- 
    (hkp://keys.openpgp.org))

PR #760, opened 2022-12-23
No tracking: remove revolvermaps

pjotrp:
I am removing tracking links.


zsloan (approved):
Seems fine to me; don't see any reason this would cause a problem.

PR #770, opened 2023-03-01
Feature/autocomplete global search

Alexanderlacuna:
- adds autocomplete feature on global search
-  enables users to see their previous history on global search
**WIP**


Alexanderlacuna:
@zsloan  please test this out   recent search and autocomplete feature on global search

PR #777, opened 2023-04-18
metadata &textfile generator script

Alexanderlacuna:

### DIFF
--- a/wqflask/scripts/textfiles_generator.py
+++ b/wqflask/scripts/textfiles_generator.py
@@ -1,4 +1,3 @@
-
 # database connection
 import contextlib
 import pickle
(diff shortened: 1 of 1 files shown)


Alexanderlacuna:
@BonfaceKilz  @zsloan  could you help in reviewing this

Alexanderlacuna:
you can use this branch for testing:https://github.com/genenetwork/genenetwork2/pull/780

zsloan:
I'm testing this now (btw sorry about the delay on that - in the future feel free to just bug me over e-mail about stuff like this), and I think it's working correctly because it's been running a long time without errors.

How do I test that it's run correctly? It's still running, but the directory I think it's writing to (in this case /tmp/Probesets) hasn't been updated since soon after I ran the script (which might be normal, but not sure).

zsloan:
After running for a while, I got the following error (which seems to be due to hitting some size limit and is likely due to the LMDB installation on Tux01 or something rather than anything in the script):
lmdb.MapFullError: mdb_put: MDB_MAP_FULL: Environment mapsize limit reached

Alexanderlacuna:
> After running for a while, I got the following error (which seems to be due to hitting some size limit and is likely due to the LMDB installation on Tux01 or something rather than anything in the script): lmdb.MapFullError: mdb_put: MDB_MAP_FULL: Environment mapsize limit reached

it has to do with allocating  size of lmdb   I will fix this ....Hope this is not occurring for --metadata  they are quite small.

zsloan:
Ah thanks, I'll try it again with that latest commit.

PR #781, opened 2023-05-11
Changes to sample data editing

zsloan:
This PR will make a couple changes to sample data editing:

- Values will be editable for samples that don't currently have values. Currently you can only edit sample values for samples that already have values; this is because it just pulls the sample data from the DB and wasn't accounting for the sample list
- It will be possible to submit sample data changes via an HTML table, instead of just as a CSV import

The corresponding GN3 PR is here - https://github.com/genenetwork/genenetwork3/pull/117

### DIFF
--- a/wqflask/wqflask/metadata_edits.py
+++ b/wqflask/wqflask/metadata_edits.py
@@ -35,6 +35,7 @@ from gn3.db import fetchall
 from gn3.db import fetchone
 from gn3.db import insert
 from gn3.db import update
+from gn3.db.datasets import retrieve_sample_list, retrieve_group_name, retrieve_trait_dataset
 from gn3.db.metadata_audit import MetadataAudit
 from gn3.db.phenotypes import Phenotype
 from gn3.db.phenotypes import Probeset
@@ -42,7 +43,7 @@ from gn3.db.phenotypes import Publication
 from gn3.db.phenotypes import PublishXRef
 from gn3.db.phenotypes import probeset_mapping
 from gn3.db.sample_data import delete_sample_data
-from gn3.db.sample_data import get_trait_csv_sample_data
+from gn3.db.sample_data import get_trait_sample_data, get_trait_csv_sample_data
 from gn3.db.sample_data import insert_sample_data
 from gn3.db.sample_data import update_sample_data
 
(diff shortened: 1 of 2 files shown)


wqflask/wqflask/metadata_edits.py
```diff
@@ -171,7 +179,7 @@ def update_phenotype(dataset_id: str, name: str):
         or ""
     )
     phenotype_id = str(data_.get("phenotype-id"))
-    if not (file_ := request.files.get("file")):
+    if not (file_ := request.files.get("file")) and data_.get('edited') == "false":
```
BonfaceKilz, line 182: *question: issue:* This seems odd to me.  Doing a string comparison the string "false".  If there's a way to do directly with a boolean value, it would make this easier to both debug and read.

BonfaceKilz (approved):
Minor comments; otherwise this LGTM.

wqflask/wqflask/metadata_edits.py
```diff
@@ -705,3 +727,26 @@ def approve_data(resource_id: str, file_name: str):
             "warning",
         )
     return redirect(url_for("metadata_edit.list_diffs"))
+
+def create_delta_csv(base_csv, form_data, sample_list):
+    base_csv_lines = base_csv.split("\n")
+    delta_csv_lines = [base_csv_lines[0]]
+
+    for line in base_csv_lines[1:]:
```
jgarte, line 735: Hi @zsloan @BonfaceKilz, maybe there's a way to code this that doesn't have a high [cyclomatic](https://scribe.rip/@commbigo/python-code-complexity-metrics-e6c87646c1c2) complexity?

Otherwise, #781 LGTM

Feel free to ignore if you think this is a nitpick...

zsloan:
I also really don't like how those nested for loops work, but I'm not sure
how else to do them. The original version of the code didn't have this
problem because it only needed to care about samples with values (which
would be the same in both the base and delta CSV files) and also had the
CSV directly submitted instead of specific values being submitted through a
form.

Basically what needs to happen is for a CSV string to be written that is
essentially the base_csv but with values substituted when they were edited
(what it's getting from the form). So it goes through each line and checks
if edits of the value, error, or n_cases were submitted with the form.

It could maybe be made a bit more simple by converting all the edits into a
single JSON value either before or after being submitted with the form,
though this would also require more code and it would still need to check
"do edits exist" for each sample/row. I think this would only circumvent
the "if form_data.get()" checks.

PR #795, opened 2023-06-22
Decouple database connection from environment

fredmanglis:
#### Description

Commit f259a48c659b6e81ad51ca3d50128737199024c3 decouples the `database_connection(...)` function from the application and its environment by adding an argument allowing the passing of the database URI as an argument.

Commit 1b8b659312059fc184f96b2708104fdaf0dc812b updates all the calls to `database_connection(...)` throughout the application.


zsloan:
I'm the process of looking at this right now, though it seems fine so far (and the nature of the changes seem like any potential issues would pop up pretty quickly, since it's a change to how it calls the database_connection).

PR #800, opened 2023-07-26
Roll back to python 3.9 in .guix-channel

BonfaceKilz:


jgarte:
Should upstream support multiple CPython versions?

BonfaceKilz:
jgart ***@***.***> aliandika:

> Should upstream support multiple CPython versions?
>
Nope.  This is unfeasible, because you'd have to
also support multiple python versions as well.
Using a specific commit that has everything you
want is more straightforward.

> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/genenetwork/genenetwork2/pull/800#issuecomment-1651787760
> You are receiving this because you modified the open/close state.
>
> Message ID: ***@***.***>

-- 
    (hkp://keys.openpgp.org))

PR #804, opened 2023-09-05
Remove escape/mescape from do_search.py

zsloan:
This PR removes all uses of escape/mescape to make the code work with Python 3.10

### DIFF
--- a/wqflask/wqflask/do_search.py
+++ b/wqflask/wqflask/do_search.py
@@ -5,7 +5,6 @@ import string
 
 from wqflask.database import database_connection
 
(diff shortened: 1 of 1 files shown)


zsloan:
Closing this, because I just did the escape part in a separate commit and Bonface dealt with mescape already

BonfaceKilz:
zsloan ***@***.***> aliandika:

> Closing this, because I just did the escape part in a separate commit and Bonface dealt with mescape already

Thanks!

-- 
    (hkp://keys.openpgp.org))

PR #808, opened 2023-10-05
UI/rdf frontend

BonfaceKilz:
#### Description

This PR adds a new front-end for the RDF work.  Of importance to the rest of the GN eco-system is that it's a working POC of integrating HTMX to do some interactive displays.  Hopefully, this PR make as good case for replacing heavy javascript to HTMX.


BonfaceKilz:
I will go ahead and merge this PR so that people I can make RFCs for people to comment on.

PR #830, opened 2024-01-26
Json decode error checking

fredmanglis:
#### Description

Commit fc9edc266c88ebb8a859bd804f4cd817a3df04e8 is some  changes I did while investigating ProbeSet and their metadata, and forgot to commit. It just fixes minor issues that were preventing the code from running successfully.

Commit 0491193305fa8065345076f3e470c0125294df4a is the main reason for this pull request. I provide better messaging, and a PoC in linking UI error messages to the related detailed messages in the logs. WDYT?

### DIFF
--- a/gn2/maintenance/generate_probesetfreeze_file.py
+++ b/gn2/maintenance/generate_probesetfreeze_file.py
@@ -27,7 +27,7 @@ def get_strains(cursor):
                       where Strain.Id = StrainXRef.StrainId and
                             StrainXRef.InbredSetId = InbredSet.Id
                             and InbredSet.Name=%s;
-                """, "BXD")
+                """, ("BXD",))
 
     strains = [strain[0] for strain in cursor.fetchall()]
     print("strains:", pf(strains))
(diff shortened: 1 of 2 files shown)


arunisaac:
I have a comment about the surrounding code and the overall approach,
rather than about your specific commit. IIUC, you are extracting some
kind of error information from the API request and logging it. The
better approach may be to simply raise this as an exception so that some
exception handler somewhere down the line can handle it.

Here's the larger philosophy. The exception system of a programming
language is meant to handle errors, and is well-suited to this
purpose. But, we use different backends connected over an API and break
the exception system. Our goal must be to bridge the API gap so that
exceptions can cross over. Thus, we enable the exception system to do
the job it was always meant to do.

The other idea in your commit is to use UUIDs to uniquely identify
specific error instances. This is clever and quite novel, but it
probably belongs in a global error handler somewhere else.

fredmanglis:
This is superseded by https://github.com/genenetwork/genenetwork2/pull/831

No more work will be done here.

PR #855, opened 2024-08-06
Gn2 Fix profiles pages dashboard

jnduli:
# Description
we have some bugs related to the profiles dashboard fixed here:

1. We don't correcly pass in the list of privileges so the [jinja templates checks](https://github.com/jnduli/genenetwork2/blob/dd7268bc0c2d841779ba488b13ca0b1f0e9ea6bc/gn2/wqflask/templates/oauth2/profile_nav.html#L26) for these will always fail
2. Searching for data to link queries gn-auth, but we don't pass in a token, resulting in a failure during search.

## Tested

`Linked Data` and `Masquerade As` are not visible in profiles page:
image

Seach now works in the `Link mrna page`:
image

### DIFF
--- a/gn2/wqflask/oauth2/ui.py
+++ b/gn2/wqflask/oauth2/ui.py
@@ -1,11 +1,7 @@
 """UI utilities"""
-from flask import session, render_template
-
-from gn2.wqflask.oauth2 import session
+from flask import render_template
 
 from .client import oauth2_get
-from .client import user_logged_in
-from .request_utils import process_error
 
 def render_ui(templatepath: str, **kwargs):
     """Handle repetitive UI rendering stuff."""
@@ -15,8 +11,7 @@ def render_ui(templatepath: str, **kwargs):
                 lambda _err: roles, lambda auth_roles: auth_roles)
     user_privileges = tuple(
         privilege["privilege_id"] for role in roles for privilege in role["privileges"])
-    user_token = session.user_token().either(lambda _err: "", lambda token: token["access_token"])
     kwargs = {
-            **kwargs, "roles": roles, "user_privileges": user_privileges, "bearer_token": user_token
+            **kwargs, "roles": roles, "user_privileges": user_privileges,
     }
     return render_template(templatepath, **kwargs)
--- a/gn2/wqflask/oauth2/ui.py
+++ b/gn2/wqflask/oauth2/ui.py
@@ -1,6 +1,8 @@
 """UI utilities"""
 from flask import session, render_template
 
+from gn2.wqflask.oauth2 import session
+
 from .client import oauth2_get
 from .client import user_logged_in
 from .request_utils import process_error
@@ -13,7 +15,8 @@ def render_ui(templatepath: str, **kwargs):
                 lambda _err: roles, lambda auth_roles: auth_roles)
     user_privileges = tuple(
         privilege["privilege_id"] for role in roles for privilege in role["privileges"])
+    user_token = session.user_token().either(lambda _err: "", lambda token: token["access_token"])
     kwargs = {
-        **kwargs, "roles": roles, "user_privileges": user_privileges
+            **kwargs, "roles": roles, "user_privileges": user_privileges, "bearer_token": user_token
     }
     return render_template(templatepath, **kwargs)
--- a/gn2/wqflask/oauth2/ui.py
+++ b/gn2/wqflask/oauth2/ui.py
@@ -8,8 +8,11 @@ from .request_utils import process_error
 def render_ui(templatepath: str, **kwargs):
     """Handle repetitive UI rendering stuff."""
     roles = kwargs.get("roles", tuple()) # Get roles
+    if not roles:
+        roles = oauth2_get("auth/system/roles").either(
+                lambda _err: roles, lambda auth_roles: auth_roles)
     user_privileges = tuple(
-        privilege for role in roles for privilege in role["privileges"])
+        privilege["privilege_id"] for role in roles for privilege in role["privileges"])
     kwargs = {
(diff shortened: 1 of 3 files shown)


gn2/wqflask/templates/oauth2/data-list-mrna.html
```diff
@@ -95,6 +95,8 @@
 	  action="{{search_uri}}"
 	  method="POST">
       <legend>Search: mRNA Assay</legend>
+
+      <input type="hidden" id="bearer_token" name="bearer_token" value="{{bearer_token}}" />
```
fredmanglis: @jnduli Oof! This induced a panic in me. We do not want to expose the bearer token to the user like this.

Here, the token is sent to the end user, rather than just a session ID, and thus could be stolen and the malicious actor could then pretend to be our user and mess things up.

Look at how I have done most of the other requests: we do not hit the auth-server with JS directly, rather we have:

```example
                                      + <--> GN3
Javascript <--> Browser <--> GN2 <--> +
                                      + <--> gn-auth
```

I would rather you "duplicate" the endpoint on GN2 which will then use the token to pass on the information to gn-auth.

gn2/wqflask/oauth2/ui.py
```diff
@@ -8,8 +8,11 @@
 def render_ui(templatepath: str, **kwargs):
     """Handle repetitive UI rendering stuff."""
     roles = kwargs.get("roles", tuple()) # Get roles
+    if not roles:
+        roles = oauth2_get("auth/system/roles").either(
+                lambda _err: roles, lambda auth_roles: auth_roles)
     user_privileges = tuple(
-        privilege for role in roles for privilege in role["privileges"])
+        privilege["privilege_id"] for role in roles for privilege in role["privileges"])
```
fredmanglis, line 17: Looks good. This can be merged.

fredmanglis (changes requested):

gn2/wqflask/static/new/javascript/auth/search_mrna.js
```diff
@@ -15,12 +15,17 @@ function search_mrna() {
     selected = JSON.parse(document.getElementById(
 	"tbl-link").getAttribute("data-datasets"));
     species_name = document.getElementById("txt-species-name").value
+    bearer_token = document.getElementById("bearer_token").value
     search_endpoint = "/auth/data/mrna/search"
```
fredmanglis: @jnduli Rather than expose the bearer token like this, use [this endpoint in GN2](https://github.com/genenetwork/genenetwork2/blob/0998033d0a7ea26ed96b00a360a334bae6de8c55/gn2/wqflask/oauth2/data.py#L107)

jnduli: Done

gn2/wqflask/templates/oauth2/data-list-mrna.html
jnduli: Ah, I see. So the ideal workflow would be:

```mermaid
sequenceDiagram
    Actor User
    par login_flow
        User->>GN2: login
        GN2->>GN-auth: login
        GN-auth->>GN2: JWT token
        GN2->>User: SessionId
    and Search Mrna Data to Link
        User->>GN2: SessionId + search_mrna
        GN2->>GN-auth: search_mrna + token
        GN-auth->>GN2: search_result
        GN2->>User: search_resullt
    end
```

why don't we want to have the tokens in the front end? I'd assumed that using tokens ideally supports this where we should be able to call `gn-auth` endpoints directly from any clients since the tokens have expiry times.

For the token getting stolen, I also think the session id also provides such a risk. Have we added any extra measures to ensure this doesn't happen? (e.g. linking the users ip_address to the session id)

fredmanglis: @jnduli 

> ... So the ideal workflow would be ...

Yes.

> ... why don't we want to have the tokens in the front end? ...

Two reasons:
- The tokens are really short-lived (300 seconds), so there's a very good chance that the token will have expired by the time you get your JS to run and do the requests. The GN2 backend knows to be able to refresh the tokens before doing any requests to the API server. If we expose the token to the browser, we'll need to then implement yet another client in JS to handle the renewal of the tokens.
- Technically, there is nothing wrong with providing a token to the user's browser, but it should be presented in a way that keeps it safe from XSS and/or CSRF attacks. Putting it in the html document makes it so that any JS can easily read the DOM and parse out the token for nefarious purposes. If you must expose the token, consider putting it in [a protected cookie](https://stackoverflow.com/questions/27067251/where-to-store-jwt-in-browser-how-to-protect-against-csrf) that mitigates some of the damage.

> For the token getting stolen, I also think the session id also provides such a risk.

Well, yes and no. The token allows one to directly access the API server. The session ID forces them to go through GN2.

i.e. with the token, you could do something like:

```sh
curl -H 'Authorization: Bearer <your-stolen-token-here>' -H 'application/json' -XPOST https://gn3.endpoint.on.genenetwork.org/ - d ...
```

> Have we added any extra measures to ensure this doesn't happen? (e.g. linking the users ip_address to the session id)

As is the case with these things, I [have left over some space to have this implemented](https://github.com/genenetwork/genenetwork2/blob/0998033d0a7ea26ed96b00a360a334bae6de8c55/gn2/wqflask/oauth2/session.py#L22-L23) but did not actually get around to it. It is not that simple, since a user, could, for example, in one session, log in on their house WIFI, travel to the  office, go to the airport, fly to a different country.

It's still the same computer, and same session, but the IP address will probably flip around like a flag in the wind.

----

TIL: One can build up diagrams in a dot-like syntax on Github: nice!

PR #857, opened 2024-08-26
feat: add template for editting wiki

jnduli:
#### Description
Add UI pages for editting wiki content. 

## Testing

Screenshot:
image

## Next up

- Ensure saving functionality works

### DIFF
--- a/gn2/wqflask/views.py
+++ b/gn2/wqflask/views.py
@@ -1494,31 +1494,46 @@ def approve_reject_diff() -> Response:
                                 diff_id=form["diff_id"]))
 
 
-@app.route("/wiki/<int:comment_id>/edit")
+@app.route("/wiki/<int:comment_id>/edit", methods=["GET", "POST"])
 def edit_wiki(comment_id: int):
     """fetch generif metadata from gn3 and display it"""
     # FIXME: better error handling
-    last_wiki_content = (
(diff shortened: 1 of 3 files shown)


gn2/wqflask/views.py
```diff
@@ -1490,3 +1492,48 @@ def __success__(results):
...
+        )
+
+        species_dict = (
+            monad_requests.get(urljoin(GN3_LOCAL_URL, "/api/metadata/wiki/species"))
+            .either(lambda err: err.raise_for_status(), lambda x: x.json())
+        )
+        categories = (
+            monad_requests.get(urljoin(GN3_LOCAL_URL, "/api/metadata/wiki/categories"))
+            .either(lambda err: err.raise_for_status(), lambda x: list(x.json().keys()))
+        )
```
BonfaceKilz: *thought: question:* I'm not a big fan of monad requests.  Is there a good reason for using it?

BonfaceKilz (approved):

gn2/wqflask/views.py
BonfaceKilz: *suggestion:* Monad requests makes things un-necessarily complex.  Please remove.

jnduli: Thanks. I've dropped this and used the `requests` library instead. I've also realized the module may be buggy since it treats any other 2XX requests as errors except 200.

BonfaceKilz (approved):

PR #858, opened 2024-08-27
UI/fetch rif using recent apis

BonfaceKilz:
#### Description

This PR updates the GeneRIF page in GN2 to use the new GN3 APIs.  It also makes the page similar to how the page in GN1 looks like.

Screenshot:
image

Todo:
- Add NCBI generif content.


gn2/wqflask/templates/genewiki.html
```diff
@@ -0,0 +1,31 @@
+{% extends "base.html" %}
+{% block title %}GeneWiki Entry for {{ symbol }}{% endblock %}
+{% block css %}<style></style>{% endblock %}
+{% block content %}
+    <div class = "container">
```
jnduli: I think we may need a `flash_me` to handle flash requests

gn2/wqflask/templates/genewiki.html
jnduli, line 1: Can we move this to a folder within templates? That's what I did in [this PR](https://github.com/genenetwork/genenetwork2/pull/857/files#diff-4c8db332bcf0eeba6abd2c859c3629ef8e596f413cdf9cb64aa1e4d0ef5f34ba). Main advantage would be that we consolidate all resources related to wiki/rif editting.

gn2/wqflask/templates/genewiki.html
```diff
@@ -0,0 +1,31 @@
...
+            <strong>GeneNetwork:</strong>
+        </h5>
+        {% if wiki %}
+            <ol>
+                {% for entry in wiki %}
+                    <li>
+                        {{ entry.get("comment") }}
+                        {% if entry.get("web_url") %}
+                            <sup> <small> <a href = "{{ entry.web_url }}" target = "_blank"> <span class = "glyphicon glyphicon-globe" aria-hidden = "true"></span>web</a></small></sup>
+                        {% endif %}
```
jnduli: can we add the history and edit buttons here?

gn2/wqflask/views.py
```diff
@@ -1219,19 +1220,19 @@ def display_diffs_users():
                            files=files)
 
 
-@app.route("/genewiki/<symbol>")
-def display_generif_page(symbol):
+@app.route("/genewiki/<string:symbol>")
```
jnduli, line 1223: what's your preference for the routes to use? The one I chose is `metadata/wiki/comment_id/edit`

gn2/wqflask/views.py
```diff
@@ -1219,19 +1220,19 @@ def display_diffs_users():
...
+def display_genewiki_page(symbol: str):
     """Fetch GeneRIF metadata from GN3 and display it"""
-    entries = requests.get(
+    wiki = requests.get(
```
jnduli: maybe handle errors here? You could call `wiki.raise_for_status()` before calling the `.json` method

gn2/wqflask/templates/genewiki.html
```diff
@@ -0,0 +1,31 @@
+{% extends "base.html" %}
+{% block title %}GeneWiki Entry for {{ symbol }}{% endblock %}
+{% block css %}<style></style>{% endblock %}
```
jnduli: do we need this block if we don't add custom styles?

gn2/wqflask/templates/genewiki.html
BonfaceKilz: Let's add this in another PR.  Let's introduce said buttons when we weave in the functionality---easier on the user.

gn2/wqflask/views.py
BonfaceKilz: Let's leave this as is.

jnduli (approved):

PR #863, opened 2024-08-30
Chores/merge gnqa systems

Alexanderlacuna:
* This` PR merges changes from the gnqa-system branch to testing
* fixes merge conflict


Alexanderlacuna:
@jnduli @BonfaceKilz  Created the branch to merge to testing for gn2.

gn2/wqflask/templates/gnqa_answer.html
```diff
@@ -93,20 +93,27 @@ <h3><b>{{ reference.pubmed[0].get('title') }}:</b></h3>
 {% block js %}
 
 <script>
+  function updateRatingHandler(target, responseObj, args){
+    let {status, response} = responseObj.xhr
+    if (status==200 && args == "upvote"){
```
jnduli: minor: `status == 200` (extra spaces)

gn2/wqflask/templates/gnqa_answer.html
```diff
@@ -93,20 +93,27 @@ <h3><b>{{ reference.pubmed[0].get('title') }}:</b></h3>
...
 var answer = {{ answer|tojson }}
 var {task_id} = {{ task_id|tojson }}
-  htmx.on("#upvote", "click", function(evt){
+htmx.on("#upvote", "click", function(evt){
       vote_count = htmx.find(".btn-success") ? 0 : 1
-      htmx.ajax("POST", `/gnqna/rating/${task_id}/${vote_count}`, {target: "#rate", swap:"innerHTML",values: {'query': query, 'answer': answer}}).then(()=>{
-	  htmx.toggleClass(htmx.find('#upvote'), 'btn-success');
-	  htmx.removeClass(htmx.find("#downvote"), "btn-danger");
-})});
+    htmx.ajax("POST", `/gnqna/rating/${task_id}/${vote_count}`, {target: "#rate", handler: (target,obj)=> updateRatingHandler(target,obj,"upvote"), swap:"innerHTML",values: {'query': query, 'answer': answer}})});
```
jnduli: can we have these htmx calls span multiple lines. This helps with code-readablity e.g.
```javascript
    htmx.ajax("POST", `/gnqna/rating/${task_id}/${vote_count}`, 
          { target: "#rate",
             handler: (target, obj) => updateRatingHandler(target, obj, "upvote"),
             swap: "innerHTML",
             values: { 'query': query,
                            'answer': answer
                          }
           }
    );
```
Also I think we have too many braces here.

gn2/wqflask/templates/gnqa_answer.html
```diff
@@ -93,20 +93,27 @@ <h3><b>{{ reference.pubmed[0].get('title') }}:</b></h3>
...
   htmx.on("#downvote", "click", function(evt){
       vote_count = htmx.find(".btn-danger") ? 0 : -1
-      htmx.ajax("POST", `/gnqna/rating/${task_id}/${vote_count}`, {target: "#rate", swap:"innerHTML",values: {'query': query, 'answer': answer}}).then(()=>{
-      htmx.toggleClass(htmx.find('#downvote'), 'btn-danger');
-      htmx.removeClass(htmx.find("#upvote"), "btn-success")
-      })});
+      htmx.ajax("POST", `/gnqna/rating/${task_id}/${vote_count}`, {target: "#rate",handler: (target,obj)=> updateRatingHandler(target,obj,"downvote") , swap:"innerHTML",values: {'query': query, 'answer': answer}})});
```
jnduli: ditto: https://github.com/genenetwork/genenetwork2/pull/863/files#r1740539125

gn2/wqflask/templates/gnqa_search_history.html
```diff
@@ -1,42 +1,52 @@
...
+	<ul class="list-group list-group-flush" style="overflow-y:scroll">
+    {% for item in  prev_queries %}
+    <li class="row list-group-item">
+      	<input name="" type="checkbox"   value="{{item['task_id']}}" class="col-sm-1" style="height: 20px;
+  width: 20px;">
 	<div class="col-sm-10">
-	  {% for id,val in  record.items() %}
 	    <button
-	      hx-get="/gnqna/hist/search/{{id}}"
+	      hx-get="/gnqna/hist?query={{item['query']}}&search_term={{item['task_id']}}"
```
jnduli: I think jinja recommends having extra spaces after `{{` and before `}}` e.g. `{{ item['query'] }}`. Could we change all jinja refs to use this instead? [ref](https://jinja.palletsprojects.com/en/3.1.x/templates/)

gn2/wqflask/templates/gnqa_search_history.html
```diff
@@ -1,42 +1,52 @@
...
 	      hx-swap="innerHTML"
 	      hx-trigger= "click"
 	      data-bs-toggle="tooltip"
 	      data-bs-placement="left"
-	      title="/gnqna/hist/search?{{id}}"
+	      title="/gnqna/hist?query={{item['query']}}&search_term={{item['task_id']}}"
 	      style="background:transparent;border:none;cursor:pointer"
 	      >
-	      <b class="text-info">{{val}} </b>
+	      <p class="text-info">{{item.get('query')}} </p>
```
jnduli: minor: why use `item.get('query')` here but in previous calls we use `item['query']`?

gn2/wqflask/templates/gnqa_search_history.html
```diff
@@ -1,42 +1,52 @@
...
+    {% endfor %}
+    </ul>
    </div>
   </div>
   </div>
- </section>
+</section>
+<script>
+  htmx.on("#delete-btn", "click", function(evt){
+      htmx.ajax("DELETE","/gnqna/hist", {target: "#search-hist","swap" :"outerHTML",
```
jnduli: :thinking: I know flask supports `url_for`to determine the end points we call. Do you know which we prefer in gn?

gn2/wqflask/views.py
```diff
@@ -300,14 +287,7 @@ def __success__(resp):
                 error_page, __success__)
         except Exception as error:
             return flask.jsonify({"error": str(error)})
```
jnduli: error type missing i.e. `return flask.jsonify({"error": str(error)}), 500`, otherwise we'll assume 200

Also, it seems we may be returning different types here i.e. if it succeeds we render_template, but on failure we return json. Can we also `render_template` with the error message?

gn2/wqflask/views.py
```diff
@@ -343,34 +323,35 @@ def commit_gn_doc():
...
-@app.route("/gnqna/hist/search/<search_term>", methods=["GET"])
+@app.route("/gnqna/hist", methods=["GET", "DELETE"])
 @require_oauth2
-def fetch_hist_records(search_term):
+def get_gnqa_history():
+    def _error_(resp):
+        return render_template("gnqa_errors.html",
+                               **{"status_code": resp.status_code,
+                                  **resp.json()})
     token = session_info()["user"]["token"].either(
```
jnduli, line 339: how do we handle the situation when a token results in an err? I expected some exception or we render an error page. Here, we'd pass the token this error to the requests_call.

gn2/wqflask/views.py
```diff
@@ -343,34 +323,35 @@ def commit_gn_doc():
...
         lambda err: err, lambda tok: tok["access_token"])
+    if request.method == "DELETE":
+        monad_requests.delete(urljoin(GN3_LOCAL_URL, "/api/llm/history"),
+                            json=dict(request.form),
```
jnduli: I thought delete requests shouldn't take in the json parameter. I don't see us explicitly define this in the `monad_requests` functions the same as we did with `put` and `post`. Also see [this comment](https://stackoverflow.com/a/32264745) for why we want to avoid this with deletes

gn2/wqflask/views.py
```diff
@@ -343,34 +323,35 @@ def commit_gn_doc():
...
+                            headers={
+                                 "Authorization": f"Bearer {token}"
+                            }
+                            ).either(
+                   _error_, lambda x: x.json())
     response = monad_requests.get(urljoin(GN3_LOCAL_URL,
-                                          f"/api/llm/history/{search_term}"),
+                 (f"/api/llm/history?search_term={request.args.get('search_term')}"
```
jnduli: can we split the ifs here to multiple lines to help with readabilty? Something like:
```
response_url = urljoin(GN3_LOCAL_URL, "/api/llm/history")
search_term = request.args.get("search_term")
if search_term:
  response_url = f"{response_url}/{search_term}
```

gn2/wqflask/views.py
```diff
@@ -343,34 +323,35 @@ def commit_gn_doc():
...
+                  if request.args.get("search_term") else "/api/llm/history")),
                                   headers={
         "Authorization": f"Bearer {token}"
     }
     ).then(lambda resp: resp).either(
-        lambda x:  x.json(), lambda x: x.json())
-    return render_template("gnqa_answer.html", **response)
+        _error_, lambda x: x.json())
```
jnduli: This looks like a bug. It we encounter an error, the value for response is the `__error__` function. How do we then handle this?

gn2/wqflask/templates/gnqa_search_history.html
Alexanderlacuna: issue resolved by running djlint for gnqa files see commit:
https://github.com/genenetwork/genenetwork2/pull/863/commits/2c13a6224aa1fe7cd223a2c9cde237758ecfee7d  and https://github.com/genenetwork/genenetwork2/pull/863/commits/003837b2e3deae96f912b0912d8000521d9ef9be

gn2/wqflask/templates/gnqa_search_history.html
Alexanderlacuna: NO I dont think there is a standard for genenetwork. cc @BonfaceKilz

gn2/wqflask/views.py
Alexanderlacuna: see this: https://requests.readthedocs.io/en/latest/_modules/requests/api/#delete   this is the  how I defined 
`it` for monads requests:https://github.com/genenetwork/genenetwork2/pull/863/commits/6db49002d4d2e69fcf4fdd6be6aceeea7b95664f

gn2/wqflask/templates/gnqa_answer.html
Alexanderlacuna: this commit fixes this: https://github.com/genenetwork/genenetwork2/pull/863/commits/f3843170bb0fe4cca17b529be61fc6ae438acfb6, On the issue of having too many braces I dont think
there is an alternative I have removed single line braces to improve readabilty.

gn2/wqflask/views.py
Alexanderlacuna: issue fixed by https://github.com/genenetwork/genenetwork2/pull/863/commits/5c1fe9fdd321998bdc3edec02a356a7d00401ebf

gn2/wqflask/templates/gnqa_search_history.html
Alexanderlacuna: This is no clear reason for this. I have made these changes for consistency: 
https://github.com/genenetwork/genenetwork2/pull/863/commits/80211e9d91d3e06ba62e1b2c08e94ae2f1d98c79

gn2/wqflask/views.py
Alexanderlacuna: Actually this exception is  unreachable the general exception handler can handle this:
todo: remove

gn2/wqflask/views.py
Alexanderlacuna: This is more of an auth question: is it possible to have a logged in user: without a token we are  already
using the @require_oauth2 decorator. Though am not sure about this.

gn2/wqflask/templates/gnqa_answer.html
Alexanderlacuna: fixed.

gn2/wqflask/templates/gnqa_answer.html
Alexanderlacuna: fixed.

gn2/wqflask/views.py
Alexanderlacuna: This commit fixes this issue: https://github.com/genenetwork/genenetwork2/pull/863/commits/eed64341c758e7f39b8f9ff2d7aee46f0fccd630

gn2/wqflask/templates/gnqa_answer.html
jnduli: thanks. I've seen the calls start from htmx.on .... so that's where the extra braces I saw came from.

gn2/wqflask/views.py
jnduli: ah, I see. So ideally, we should never end up with an invalid token. I'm not too sure about how to handle this too. Can we add a TODO comment here?

jnduli (approved):

PR #864, opened 2024-09-03
Add comment history page.

BonfaceKilz:
#### Description

This PR creates a comment history page:

An entry with the history link:
image

The history page:
image

### DIFF
--- a/gn2/wqflask/templates/wiki/history.html
+++ b/gn2/wqflask/templates/wiki/history.html
@@ -14,42 +14,35 @@
                             <th>Gene Symbol:</th>
                             <td>{{ most_recent.symbol }}</td>
                         </tr>
-                        {% if most_recent.pubmed_ids %}
-                            <tr>
-                                <th>PubMed IDs:</th>
-                                <td>
-                                    {% for id in most_recent.pubmed_ids %}
-                                        <a href="http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids={{ id }}&dopt=Abstract">{{ id }}</a>&nbsp;
-                                    {% endfor %}
-                                </td>
-                            </tr>
-                        {% endif %}
-                        {% if most_recent.web_url %}
-                            <tr>
-                                <th>Web URL:</th>
-                                <td>{{ most_recent.web_url }}</td>
-                            </tr>
-                        {% endif %}
+                        <tr>
+                            <th>PubMed IDs:</th>
+                            <td>
+                                {% for id in most_recent.pubmed_ids %}
+                                    <a href="http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids={{ id }}&dopt=Abstract">{{ id }}</a>&nbsp;
+                                {% endfor %}
+                                {% if not most_recent.pubmed_ids %}Not Available{% endif %}
+                            </td>
+                        </tr>
+                        <tr>
+                            <th>Web URL:</th>
+                            <td>{{ most_recent.web_url or "Not Available" }}</td>
+                        </tr>
                         <tr>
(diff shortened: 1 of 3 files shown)


gn2/wqflask/templates/wiki/history.html
```diff
@@ -0,0 +1,141 @@
...
+            margin-left: 1em;
+            max-width: 60em !important;
+        }
+    </style>
+{% endblock %}
+{% block content %}
+    {{ flash_me() }}
+    <div class = "container container-fluid">
+        <h2>GeneWiki Entry History</h2>
+        {% if most_recent %}
```
jnduli: I dont think we need this if

gn2/wqflask/templates/wiki/history.html
```diff
@@ -0,0 +1,141 @@
...
+                            <td>{{ most_recent["reason"] }}</td>
+                        </tr>
+                    {% endif %}
+                </tbody>
+            </table>
+        {% endif %}
+        <h3>
+            <strong>Previous Version:</strong>
+        </h3>
+        {% if previous_versions %}
```
jnduli: I don't think we need this if

gn2/wqflask/templates/wiki/history.html
```diff
@@ -0,0 +1,141 @@
+{% extends "base.html" %}
+{% block css %}
+    <style type="text/css">
+        h3 {
+            color: #336699;
```
jnduli: perhaps use bootstrap's contextual colors instead? https://getbootstrap.com/docs/3.3/css/#helper-classes-colors

gn2/wqflask/templates/wiki/history.html
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        table tbody th {
```
jnduli: can we try to use grid-adjacent elements here instead?  e.g. we can add the table into a div to enforce this width, something like:
```
<div class="row">
   <div class="col-sm-9">
   <table ...>
  </div>
</div>
```

gn2/wqflask/templates/wiki/history.html
```diff
@@ -0,0 +1,141 @@
...
+            <strong>Previous Version:</strong>
+        </h3>
+        {% if previous_versions %}
+            {% for version in previous_versions %}
+                <table class="table table-responsive table-bordered">
+                    <tr>
+                        <th>Gene Symbol:</th>
+                        <td>{{ version["symbol"] }}</td>
+                    </tr>
+                    {% if version["species"] %}
```
jnduli: I think it would be better to display the rows even if they didn't exist and provide a sane sentinel value e.g. `{{ version.get("species", "not available") }}` which serves two functions:

- makes the template easier to read and parse since we drop all the ifs
- shows the user that this value doesn't exists.

This also applies to mos of the other ifs

gn2/wqflask/views.py
```diff
@@ -1277,6 +1277,29 @@ def display_genewiki_page(symbol: str):
...
+    try:
+        entries = requests.get(
+            urljoin(
+                GN3_LOCAL_URL,
+                f"/api/metadata/wiki/{comment_id}/history"
+            )
+        )
+        entries.raise_for_status()
+        if entries := entries.json():
+            most_recent, previous_versions = entries[0], entries[1:]
```
jnduli: I wonder if we can have this logic in the template itself, removing the need for duplicating the two sections. Something like:
```
{% for i in range(len(entries) %}
{% if i == 0 % }
Most Recent Version
{% elif i == 1 %}
Other versions
{% endif %}
normal table calls
```

gn2/wqflask/templates/wiki/history.html
BonfaceKilz: Nice catch!

BonfaceKilz:
@jnduli I've made all the requested suggestions.  Thanks!

PR #867, opened 2024-09-12
Llm integration bug fix

jnduli:
# Description

Bug fixes for LLM integration work that includes:

- ensure `gnqa_answers` jinja templates extends `base.html`
- send `get` request instead of `put` when accessing gn3's llm search api
- provide a rough implementation for the `xapian_query_cleaner` implementation.
- add a spinner when api search is happening

# Tested

Results after search screenshot:
image

And the ai results page:
image

### DIFF
--- a/gn2/wqflask/views.py
+++ b/gn2/wqflask/views.py
@@ -324,7 +324,7 @@ def gnqna():
                     "search_url": f"/gnqna?{safe_query}",
                 }
                 return jsonify(ai_result)
-            return render_template("gnqa_answer.html", **search_result)
+            return render_template("gnqa.html", **search_result)
         else:
             return render_template("gnqa.html")
     if request.method == "POST":
--- a/gn2/wqflask/views.py
+++ b/gn2/wqflask/views.py
@@ -270,9 +270,21 @@ def gsearchtable():
 
 def clean_xapian_query(query: str) -> str:
     """ Remove filler words in xapian query
+    This is a temporary solution that works for some query. A better solution is being worked on.
     TODO: FIXME
     """
-    return query
+    xapian_prefixes = set(["author", "species", "group", "tissue", "dataset", "symbol", "description", "rif", "wiki"])
+    range_prefixes = set(["mean", "peak", "position", "peakmb", "additive", "year"])
+    final_query = []
+    for word in query.split():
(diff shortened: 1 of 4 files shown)


Alexanderlacuna (changes requested):
"I don’t think you should extend `base.html`. The `gnqa_answer` page is a subpage of `gnqa` that  extends `base.html`, which is rendered dynamically using htmx."

Alexanderlacuna:
Related to the above(context screenshot)
Question : Wasn't "See reference" supposed to be a button that opens to a  new webpage in this case the  gnqa_anwer page??

gn2/wqflask/views.py
```diff
@@ -270,9 +270,21 @@ def gsearchtable():
...
-    return query
+    xapian_prefixes = set(["author", "species", "group", "tissue", "dataset", "symbol", "description", "rif", "wiki"])
+    range_prefixes = set(["mean", "peak", "position", "peakmb", "additive", "year"])
+    final_query = []
+    for word in query.split():
+        split_word = word.split(":")
+        if len(split_word) > 0 and split_word[0].lower() in xapian_prefixes:
+            final_query.append(split_word[1])
+            continue
+        if split_word[0].lower() in range_prefixes:
```
Alexanderlacuna, line 284: this  looks good

jnduli:
> "I don’t think you should extend `base.html`. The `gnqa_answer` page is a subpage of `gnqa` that extends `base.html`, which is rendered dynamically using htmx."

Thanks for catching this. I've fixed this and included `gnqa_answer` in the `gnqa.html` page when we send a GET request.

jnduli:
> Related to the above(context screenshot) Question : Wasn't "See reference" supposed to be a button that opens to a new webpage in this case the gnqa_anwer page??

Yeah, you're right. However when I used the `gnqa_answer.html` pages, it didn't look good since it didn't extend base.html. Right now the change I've done uses `gnqa.html` and loads up the content for this.

The `References` link still opens a new tab, but in this case it opens up the `GNQA` search page with results already pre-populated.

jnduli:
Here's how the search page looks like with GET requests:
image

Alexanderlacuna (approved):

PR #868, opened 2024-09-12
feat: implement text transformer for xapian searches.

Alexanderlacuna:
PR implements text transformer for xapian queries to feed to fahamu api.

### DIFF
--- a/gn2/wqflask/views.py
+++ b/gn2/wqflask/views.py
@@ -268,11 +268,47 @@ def gsearchtable():
 
     return flask.jsonify(current_page)
 
+
 def clean_xapian_query(query: str) -> str:
-    """ Remove filler words in xapian query
-    TODO: FIXME
     """
-    return query
+    Clean and optimize a Xapian query string by removing filler words,
+    and ensuring the query is tailored for optimal results from Fahamu.
+
+    Args:
+        query (str): The original Xapian query string.
+
+    Returns:
+        str: The cleaned and optimized query string.
+    """
+    xapian_prefixes = {
+        "author",
+        "species",
+        "group",
+        "tissue",
+        "dataset",
+        "symbol",
+        "description",
+        "rif",
+        "wiki",
+    }
+    xapian_operators = {"AND", "NOT", "OR", "XOR", "NEAR", "ADJ"}
+    range_prefixes = {"mean", "peak", "position", "peakmb", "additive", "year"}
+    query_context = ["genes"]
+    cleaned_query_parts = []
+    for token in query.split():
+        if token in xapian_operators or any(
+            prefix in token for prefix in range_prefixes if ".." in token
+        ):
+            continue
+        prefix, _, suffix = token.partition(":")
+        if prefix in xapian_prefixes:
+            query_context.insert(0, prefix)
+            cleaned_query_parts.append(f"{prefix} {suffix}")
+        else:
+            cleaned_query_parts.append(prefix)
+    cleaned_query = " ".join(cleaned_query_parts)
+    context = ",".join(query_context)
+    return f"Provide answer on {cleaned_query} context {context}"
 
 
 @app.route("/gnqna", methods=["POST", "GET"])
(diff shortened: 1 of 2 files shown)


gn2/wqflask/views.py
```diff
@@ -268,11 +268,47 @@ def gsearchtable():
...
+        "rif",
+        "wiki",
+    }
+    xapian_operators = {"AND", "NOT", "OR", "XOR", "NEAR", "ADJ"}
+    range_prefixes = {"mean", "peak", "position", "peakmb", "additive", "year"}
+    query_context = ["genes"]
+    cleaned_query_parts = []
+    for token in query.split():
+        if token in xapian_operators or any(
+            prefix in token for prefix in range_prefixes if ".." in token
```
jnduli: What conditions are we handling here? I can only think of `mean:8..` and `Hs:chr4:9930021`. These can be handled by:
```
if prefix in range_prefixes:
  continue
```
I'm not sure the `".." in token` is needed

gn2/wqflask/views.py
```diff
@@ -268,11 +268,47 @@ def gsearchtable():
...
+            continue
+        prefix, _, suffix = token.partition(":")
+        if prefix in xapian_prefixes:
+            query_context.insert(0, prefix)
+            cleaned_query_parts.append(f"{prefix} {suffix}")
+        else:
+            cleaned_query_parts.append(prefix)
+    cleaned_query = " ".join(cleaned_query_parts)
+    context = ",".join(query_context)
+    return f"Provide answer on {cleaned_query} context {context}"
```
jnduli: Question: 
1. Does this results in better results and should we somehow integrate this in the normal searches?
2. Should we add some type of formatting to the text i.e.
```bash
# if I search for:
rif:glioma species:human group:gtex_v8
# we pass in the query:
Provide answer on glioma human gtex_v8 context rif,species,group
# I thought we'd add some identifiers like
Provide answer on `glioma human gtex_v8` with context `[rif, species,group]`
```

Alexanderlacuna: 1) Yeah, I conducted some experiments on this and others. It reduces the amount of 
"No feedback response/text does not contain".... responses we get
 also provides  more information on the subject with context provided
 Ideally Fahamu should be able to pick few keywords and get a response
 but currently it doesn't If this improves the normal search also, that would be okay
2) I would need to test this.

gn2/wqflask/views.py
Alexanderlacuna: These conditions are handled where token is  either

```
XAPIAN OPERATOR .e.g AND
RANGE 
```
More on this:
From  the docs this is the range syntax is word:
 https://xapian.org/docs/queryparser.html. 
This means that they are only considered ranges if they include the range operator (`..`) as part of their token. For example, `mean:5...7` should be rejected, whereas `mean` should be accepted as a word. 
Ideally this check should be  token has : and also .. and strip the word too
I will improve on this for clarity.

zsloan:
Does it make sense to have the clean_xapian_query function in views.py? I think the idea with that file is just to contain the various routings for GN2.

That being said, I'm not sure where else would be a better place to put it, so this isn't a big deal. It's just that the file in question is already pretty long.

Alexanderlacuna:
> Does it make sense to have the clean_xapian_query function in views.py? I think the idea with that file is just to contain the various routings for GN2.
> 
> That being said, I'm not sure where else would be a better place to put it, so this isn't a big deal. It's just that the file in question is already pretty long.

I could define it  inside the `/gnqna` endpoint since it is more of a utility for one of its inputs.

gn2/wqflask/views.py
jnduli: ah, I see. Thanks for the explanation. Then can we change this to:

```
prefix, _, suffix = token.partition(":") # line already defined below
if ".." in suffix and prefix in range_prefixes:
  continue
```

jnduli (approved):

PR #871, opened 2024-09-16
feat: show log in message for search results

jnduli:
#### Description
UI improvement to show users a login message when they search.

Screenshot:

image

### DIFF
--- a/gn2/wqflask/templates/gsearch_gene.html
+++ b/gn2/wqflask/templates/gsearch_gene.html
@@ -23,6 +23,12 @@
            <i class="fa fa-spinner fa-spin fa-3x"></i>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


zsloan:
I think it might be best to not show this at all if the user isn't logged
in.

Also, most users probably won't be familiar with "LLM" as an acronym* (this
might also be an issue elsewhere, but I just noticed it here mainly because
global search is one of the more commonly used pages - most users will be
confused if they see this).

* Use of the term "LLM" to refer to stuff like GPT is pretty rare, at least
in the US (even though, in my opinion, it's far more accurate than "AI").
GeneNetwork users might be especially confused, since they might mix it up
with "LMM" - linear mixed models, the method we use for GWAS mapping.

PR #875, opened 2024-09-17
Chores/implement redirect for users 

Alexanderlacuna:
This PR add these supports:
* [] redirect users to login page if the service requires login
* [] after successful login redirect users to the correct page.
See this for the full issue description:
https://issues.genenetwork.org/issues/gn-auth/implement-redirect-on-login

### DIFF
--- a/gn2/wqflask/oauth2/checks.py
+++ b/gn2/wqflask/oauth2/checks.py
@@ -2,7 +2,7 @@
 from functools import wraps
 from urllib.parse import urljoin
 
-from flask import flash, request, redirect
+from flask import flash, request, redirect, url_for
 from authlib.integrations.requests_client import OAuth2Session
 
 from . import session
@@ -20,10 +20,21 @@ def require_oauth2(func):
     @wraps(func)
     def __token_valid__(*args, **kwargs):
         """Check that the user is logged in and their token is valid."""
-
         def __clear_session__(_no_token):
             session.clear_session_info()
-            # redirect to the login page
+            flash("You need to be logged in.", "alert-warning")
+            return redirect("/")
+
+        def __redirect_to_login__(_token):
+            """
+            Save the current user request to session then
+            redirect to the login page.
+            """
+            if request.method == "GET":
+                redirect_url = url_for(request.endpoint, **request.args)
+            else:
+                redirect_url = "/"
+            session.set_redirect_url(redirect_url, **request.args)
             return redirect(authserver_authorise_uri())
 
         def __with_token__(token):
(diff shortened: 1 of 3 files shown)


gn2/wqflask/oauth2/checks.py
```diff
@@ -12,29 +12,41 @@
...
+from .request_utils import authserver_authorise_uri
 
 
 def require_oauth2(func):
     """Decorator for ensuring user is logged in."""
     @wraps(func)
     def __token_valid__(*args, **kwargs):
         """Check that the user is logged in and their token is valid."""
-
         def __clear_session__(_no_token):
```
jnduli: drop this function?

gn2/wqflask/oauth2/checks.py
```diff
@@ -12,29 +12,41 @@
...
 
+        def __redirect_to_login__(_token):
+            """
+            Save the current user request to session then
+            redirect to the login page.
+            """
+            if request.method == "GET":
+                redirect_url = url_for(request.endpoint, **request.args)
+            else:
+                redirect_url = "/"
```
jnduli, line 32: A minor improvement would be to assume all requests have a corresponding `GET`, so we could change this to:
```
# Note: this isn't tested but writen to demonstrate the idea
try:
  redirect_url = url_for(rquest.endpoint, _method="GET", **request.args)
except BuildError:
  redirect_url = "/"
```

Here's the tests I've done for this:
```
(Pdb) url_for("gsearch_updating", _method="GET")
*** werkzeug.routing.BuildError: Could not build url for endpoint 'gsearch_updating' ('GET'). Did you mean to use methods ['OPTIONS', 'POST']?
(Pdb) url_for("gsearch_updating", _method="POST")
'/gsearch_updating'
```

The advantage is that for endpoints that support both `GET` and `POST`, we'll redirect to the page that had the form

Alexanderlacuna: Thanks, made the requested changes.

jnduli (approved):

gn2/wqflask/oauth2/session.py
```diff
@@ -118,3 +119,10 @@ def toggle_token_refreshing():
 def is_token_refreshing():
     """Returns whether the token is being refreshed or not."""
     return session_info().get("token_refreshing", False)
+
+def set_redirect_url(url):
```
fredmanglis, line 123: This looks good, but you should probably add a related `clear_redirect_url` function to clean up after yourself once the login is complete, but just before you do the redirect.

gn2/wqflask/oauth2/toplevel.py
```diff
@@ -81,7 +81,8 @@ def __success__(token):
                 "token": session.user_token(),
                 "logged_in": True
             })
-            return redirect("/")
+            redirect_url = session.session_info().get("redirect_url", "/")
+            return redirect(redirect_url)
```
fredmanglis, line 85: Clear the "redirect_url" value in the session here, just before the `return redirect(redirect_url)` statement. Don't leave the now unnecessary value in place.

fredmanglis (commented):
Other than the 2 comments regarding clearing the temporary value from the session, this looks good.

Alexanderlacuna:
@fredmanglis Thanks for the review, Well noted.

PR #876, opened 2024-09-20
Fix cases where index for '.' and ',' do not exist  for do_ai_search.

Alexanderlacuna:
PR fixes issue for do ai search function

### DIFF
--- a/gn2/wqflask/templates/gsearch_gene.html
+++ b/gn2/wqflask/templates/gsearch_gene.html
@@ -290,11 +290,13 @@
         let header_length = 40;
         let max_header_length = 180;
         while (header_length <= search_result.length) {
-          let next_dot = search_result.indexOf(".", header_length+1);
-
+            let next_dot = search_result.indexOf(".", header_length+1);
+	    if (next_dot == -1) {
+		next_dot = search_result.length+1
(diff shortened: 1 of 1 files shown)


gn2/wqflask/templates/gsearch_gene.html
```diff
@@ -290,11 +290,13 @@ <h3>GN searched for the term(s) <b>"{{ terms }}"</b> in 754 datasets and 39,765,
         let header_length = 40;
         let max_header_length = 180;
         while (header_length <= search_result.length) {
-          let next_dot = search_result.indexOf(".", header_length+1);
-
+            let next_dot = search_result.indexOf(".", header_length+1);
+	    if (next_dot == -1) {
```
jnduli, line 294: :thinking: can we instead do `next_dot = header_length`?

and then change the else block to have:

```
else if next_dot == header_length {
   break;
} else {
  header_length = next_dot
}
```

Alexanderlacuna: Rather than just terminating the check, I also want to look for a comma, which could serve as a breakpoint when `next_dot` exceeds `max_header_length`.

PR #877, opened 2024-09-23
Feature/cleanup base file

Alexanderlacuna:
This PR cleans up the base file. It includes the following changes:
* [] Move the global search to the index page
* [] Remove custom JS from the base file and add it to the index file
* [] Move the global search to the index page
* [] fix incorrect html tags
* []  formatting for base file

### DIFF
--- a/gn2/wqflask/templates/base.html
+++ b/gn2/wqflask/templates/base.html
@@ -177,7 +177,7 @@
                                     </li>
                                 </ul>
                             </li>
-                            {% if g.user_session.logged_in %}
+                            {% if logged_in() %}
                                 <li>
                                     <a href="/edit_account_settings"
                                        class="dropdown-toggle"
--- a/gn2/wqflask/templates/base.html
+++ b/gn2/wqflask/templates/base.html
@@ -33,32 +33,6 @@
          table.dataTable thead .sorting_desc {
              background-image: url({{ url_for("js", filename="DataTables/images/sort_desc_disabled.png") }});
          }
-         .global_search_input{
-             padding:9px 8px;
-             text-decoration: none;
-             border: none;
-
-             border-radius: 5px;
-         }
-         .global_search_input:focus{
-             outline: none;
-         }
-         .btn-stc {
-             padding:9px 8px;
-	     border-left:none;
-	     border-radius:0 40px 40px 0;
-	     cursor: pointer;
-	     height: 40px;
-             width: 64px;
-             margin:0;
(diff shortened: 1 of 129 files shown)


BonfaceKilz (approved):
@Alexanderlacuna I've tested out your work locally and looked at the individual commits.  Great work for the general clean-up.  We can have this merged when you are ready.  Thanks!

gn2/wqflask/templates/base.html
```diff
@@ -33,262 +46,297 @@
...
+                                    </li>
+                                    <li>
+                                        <a href="https://github.com/genenetwork/sysmaintenance">System Maintenance Code</a>
+                                    </li>
+                                    <li>
+                                        <a href="https://github.com/genenetwork/genenetwork2/blob/testing/doc/API_readme.md">REST API Documentation</a>
+                                    </li>
+                                </ul>
                             </li>
+                            {% if g.user_session.logged_in %}
```
jnduli: nitpick: `if logged_in() `? The advantage is that the function also checks if the token is valid

jnduli (changes requested):
My main concern regards losing the search button in may other pages. I've done a `git grep base.html` and there 125 files that will be affected. 

```
╰─$ git --no-pager grep "base.html" | wc -l
125
```

Here's an example scenario:
image
 
 Are we ok with this? I think if not, we could just change these pages to extend `index.html` instead.

BonfaceKilz:
@jnduli 

> My main concern regards losing the search button in may other pages. I've done a `git grep base.html` and there 125 files that will be affected.
> 
> ```
> ╰─$ git --no-pager grep "base.html" | wc -l
> 125
> ```
> 
> Here's an example scenario:
> [...]
> 
> Are we ok with this? I think if not, we could just change these pages to extend `index.html` instead.

@jnduli thanks for catching the above.

Alexanderlacuna:
@jnduli This may need clarification. CC @BonfaceKilz, CC @zsloan. Although I don’t think we need to include global search for all pages, especially those that are not related to it, such as gnqna, docs, etc., in my opinion, it should only be implemented for search-related functionalities and pages that render search results. In this PR, I have only added it for the index page and extended the functionality to the gsearch result pages.

gn2/wqflask/templates/base.html
Alexanderlacuna: Done!

jnduli (approved):

BonfaceKilz (approved):

PR #881, opened 2024-10-11
Add display page for NCBI RIF metadata.

BonfaceKilz:
#### Description

This PR implements the display page for NCBI RIF entries:

image

## Testing

1. Set up this [PR](https://github.com/genenetwork/genenetwork3/pull/200)
Note: Use the TTL file from the above PR

2. Navigate to: `/genewiki/Shh` to verify.

### DIFF
--- a/gn2/wqflask/templates/wiki/genewiki.html
+++ b/gn2/wqflask/templates/wiki/genewiki.html
@@ -64,5 +64,4 @@
             </p>
         {% endif %}
     </div>
-    <div class="container"></div>
 {% endblock %}
--- a/gn2/wqflask/templates/wiki/genewiki.html
+++ b/gn2/wqflask/templates/wiki/genewiki.html
@@ -11,26 +11,22 @@
         <h5>
             <strong>GeneNetwork:</strong>
         </h5>
-
         {% if wiki %}
-
             <ol class="list-group">
                 {% for entry in wiki %}
                     <li class="list-group-item">
                         <div class="row">
-                            <div class="col-sm-1">{{ loop.index }}.</div>
                             <div class="col-sm-6">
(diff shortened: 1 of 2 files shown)


gn2/wqflask/templates/wiki/genewiki.html
```diff
@@ -43,5 +39,30 @@ <h5>
...
+                    </li>
+                {% endfor %}
+            </ol>
+        {% else %}
+            <p class = "well">
+                <u> There are no NCBI RIF entries for this for<b> {{ symbol }}.</b></u>
+            </p>
+        {% endif %}
     </div>
+    <div class="container"></div>
```
jnduli: why do we have this empty div?

gn2/wqflask/templates/wiki/genewiki.html
```diff
@@ -43,5 +39,30 @@ <h5>
...
+                                (<a href="http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids={{ entry.pubmed_id }}&dopt=Abstract"
+   target="_blank">PubMed</a>)
+                            </div>
+                        </div>
+                    </li>
+                {% endfor %}
+            </ol>
+        {% else %}
+            <p class = "well">
+                <u> There are no NCBI RIF entries for this for<b> {{ symbol }}.</b></u>
```
jnduli, line 63: why have the </b> here? Also, should we have a starting `<b>`, [mdn ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b)

BonfaceKilz: I do this to maintain UI compatibility with GN1.

PR #882, opened 2024-10-14
GN editting UI improvements

jnduli:
#### Description
Add UI button linking to GN2 editting feature:

image

And add functionality where clicking on `edit` without login informs you to login:

image

### DIFF
--- a/gn2/wqflask/templates/show_trait_details.html
+++ b/gn2/wqflask/templates/show_trait_details.html
@@ -222,7 +222,7 @@
         <button type="button" class="btn btn-default" title="Check probe locations at UCSC" onclick="window.open('{{ UCSC_BLAT_URL }}', '_blank')">Verify</button>
         {% endif %}
         {% if this_trait.symbol != None %}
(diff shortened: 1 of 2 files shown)


gn2/wqflask/templates/show_trait_details.html
```diff
@@ -222,6 +222,7 @@
         <button type="button" class="btn btn-default" title="Check probe locations at UCSC" onclick="window.open('{{ UCSC_BLAT_URL }}', '_blank')">Verify</button>
         {% endif %}
         {% if this_trait.symbol != None %}
+	<button type="button" class="btn btn-default" title="Write or review comments about this gene" onclick="window.open('http://localhost:5012/genewiki/{{ this_trait.symbol }}', '_blank')">(GN2) GeneWiki</button>
```
Alexanderlacuna: I don't this  is okay . What will happen if the port we are currently using changes? I would  recommend you  use  a dynamic path obtained from the server.

Question: Is it okay   to expose the port that a service is running  to the client?

Alexanderlacuna (changes requested):

gn2/wqflask/templates/show_trait_details.html
jnduli: Thanks for catching this :D. You're right, this is not ok.

jnduli: Done. Comment addressed.

PR #936, opened 2026-05-05
Ignore failures fetching aliases.

fredmanglis:
In the case where fetching aliases (fetched from gn-guile) fails due to an issue with the connection, simply ignore and move on without them rather that allowing a catastrophic failure.

#### Description

This commit should allow the `show_trait` page to load, even if the aliases are not loaded due to a network error. Any other error will likely lead to a failure.

### DIFF
--- a/gn2/base/trait.py
+++ b/gn2/base/trait.py
@@ -161,14 +161,24 @@ class GeneralTrait:
 
         alias = 'Not available'
         if self.symbol:
-            human_response = requests.get(urljoin(GN_GUILE_SERVER_URL, f"gene/aliases/{self.symbol.upper()}"))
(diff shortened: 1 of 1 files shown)


zsloan:
While this will prevent an error, will it also address the issue of the page hanging while waiting on a response?

Either way I'll go ahead and merge it since it's better than the status quo.

fredmanglis:
> While this will prevent an error, will it also address the issue of the page hanging while waiting on a response?
> 

Hmm, not really... Each of the requests will wait until complete/timeout before proceeding.

We could look into doing the requests concurrently so that we can reduce the waiting time, but that comes at the expense of slightly more complicated code.

> Either way I'll go ahead and merge it since it's better than the status quo.

Thanks. Let's see whether the delays will be a problem.

